Issue #22, opened 2021-03-04
PNG restricts ICC to ICC v.2


svgeesus:
Raised by: Phil Green

> A profile is embedded in a PNG file as a chunk with type ‘iCCP’, identifying an arbitrary profile name, compression method, and the compressed profile. Alternatively, a PNG file can implicitly reference an sRGB transform in an ‘sRGB’ chunk type. Full details are given in the PNG specification at https://www.w3.org/TR/PNG/. Currently the PNG specification restricts the ICC profile to v2.

The latest versions of ICC are :

Specification ICC.1:2010-12 (Profile version 4.3.0.0) Image technology colour management - Architecture, profile format, and data structure 
http://www.color.org/icc_specs2.xalter which is also published as ISO 15076-1, 

and 

Specification ICC.2:2019 (iccMAX) Image technology colour management - Extensions to architecture, profile format, and data structure 
http://www.color.org/icc_specs2.xalter which is also published as ISO 20677-1

svgeesus:
The restriction to v.2 was not intentional, it was simply the latest version available at the time that PNG second edition was standardized.

In practice, ICC v.4 profiles are routinely embedded in PNG images.

The specification should be updated to point to the latest ICC specifications.

svgeesus:
The normative references are now:

> ICC-1, International Color Consortium, "Specification ICC.1:2010-12 (Profile version 4.3.0.0) Image technology colour management - Architecture, profile format, and data structure, available at http://www.color.org/icc_specs2.xalter also published as ISO 15076-1

> ICC-2, International Color Consortium, "Specification ICC.2:2019 (iccMAX) Image technology colour management - Extensions to architecture, profile format, and data structure, available at http://www.color.org/icc_specs2.xalter also published as ISO 20677-1

svgeesus:
I alerted Phil Green (ICC Technical Secretary) to this change, and await his confirmation that this solves the issue.

svgeesus:
> Hi Chris

> Fantastic, thanks. Good to see iccMAX supported as well as v4. I'll update the profile embedding document.

> Phil

AlexTMjugador:
The current first public working draft of the third edition of the spec does not mention ICC v2 as a normative reference, which means that iCCP chunks with ICC v2 profile data do no longer conform to the spec.

[...]

PR #34, opened 2021-11-22
Update ICC normative refs, for #22

svgeesus:
Adds ICC.1 and ICC.2 to normative references

### DIFF
--- a/index.html
+++ b/index.html
@@ -830,21 +830,25 @@ space &mdash; sRGB,</i> available at <code><a href=
 3-900-734-00-3.</p>
 
 <p class="NormRefDef"><a name="2-ICC-1">ICC-1</a>, International
-Color Consortium, "Specification ICC.1: 1998-09, File Format for
-Color Profiles", 1998, available at <code><a href=
-"http://www.color.org/">http://www.color.org/</a></code></p>
+Color Consortium, "Specification ICC.1:2010-12 (Profile version 4.3.0.0)
+Image technology colour management - Architecture, profile format, and data structure,
+available at <code><a href=
+"http://www.color.org/icc_specs2.xalter">http://www.color.org/icc_specs2.xalter</a></code></p>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -830,21 +830,26 @@ <h1><a name="2NormRefs">2 Normative references</a></h1>
...
-
-<p class="NormRefDef"><a name="2-ICC-1A">ICC-1A</a>,
-International Color Consortium, "Specification ICC.1A: 1999-04,
-Addendum 2 to ICC.1: 1998-09", 1999, available at <code><a href=
-"http://www.color.org/">http://www.color.org/</a></code></p>
+Color Consortium, "Specification ICC.1:2010-12 (Profile version 4.3.0.0)
+Image technology colour management - Architecture, profile format, and data structure,
+available at <code><a href=
+"http://www.color.org/icc_specs2.xalter">http://www.color.org/icc_specs2.xalter</a></code>
+also published as ISO 15076-1</p>
```
ProgramMax, line 837: This change has ICC-1 references (used in the iCCP & sRGB chunk definitions) pointing to the v4 ICC profiles.
But iCCP should remain v2 and the new iCCN should be v4?

I think the updated link is good. But it should probably say v2 instead of v4.

The sRGB chunk references it for the rendering intent. Maybe that is shared between v2 and v4? I'm not familiar enough here to know.

index.html
```diff
@@ -830,21 +830,26 @@ <h1><a name="2NormRefs">2 Normative references</a></h1>
...
+
+<p class="NormRefDef"><a name="2-ICC-a">ICC-2</a>,
```
ProgramMax, line 839: The link target changed from 2-ICC-1A to 2-ICC-a.

The places that point to this fragment should be updated as well.

ProgramMax (commented):
Since this is already merged, we might want to make some follow-up changes.

Issue #27, opened 2021-05-05
image/png registration points to Internet Draft


svgeesus:
The IANA [registration for `image/png`](https://www.iana.org/assignments/media-types/image/png) gives the defining specification as

> Internet-Draft draft-boutell-png-spec-04.txt

This should be updated to point to a more stable specification which didn't expire in 1996

ProgramMax:
Would it be best to point to the [W3C spec](https://www.w3.org/TR/PNG) or [ISO spec](https://www.iso.org/standard/29581.html)?

Do we know if the contacts on the IANA doc are the people we should contact about this?

ProgramMax:
I think it would be best to link to the W3C spec which is freely available.

Also, it looks like there is [good documentation for registering/updating IANA MIME types](https://www.w3.org/2020/01/registering-mediatypes).

> Note: Once a media type has been published by the IANA, the owner may request a change to its definition. The same procedure that would be appropriate for the original registration request is used to process a change request. See also Section 5.5, Change Procedures in BCP 13.

Also, from section 4.3 of that link:

> If your registration points to a dated version, request update of the registration via iana-mime-comment.

svgeesus:
Yes, this is simple to do. I have been on the IANA MIME types list since 1992 so I am familiar with how it works.
Also the whole of [Annex A](https://w3c.github.io/PNG-spec/#A-Conventions) predates the modern IANA registration template and should be replaced by a filled-in IANA template.

Related to https://github.com/w3c/PNG-spec/issues/26 as we should register the MIME type for APNG as well (with a second filled-in template)

svgeesus:
> Would it be best to point to the W3C spec or ISO spec?

The W3C spec, as you said, not only because it will be freely available but also because it will be updated sooner. And we don't yet know if the next version of PNG will have the same ISO number.

> Do we know if the contacts on the IANA doc are the people we should contact about this?

The [people listed](https://datatracker.ietf.org/doc/html/draft-boutell-png-spec-04#section-19) are (a subset of) the original PNG authors and were involved in creating the original PNG 1.0 [RFC 2083](https://datatracker.ietf.org/doc/html/rfc2083) in 1997.

svgeesus:
From RFC 6838, [6.1.  Change Procedures](https://datatracker.ietf.org/doc/html/rfc6838#section-6.1)

>   Registrations may be updated in each registry by the same mechanism
   as required for an initial registration.  In cases where the original
[...]

PR #36, opened 2021-11-22
Add IANA registration template, see #27

svgeesus:
Replaces the old Annex A (file extension, media type, mac type) with the current IANA registration template which has all that and more in a standard format.

### DIFF
--- a/index.html
+++ b/index.html
@@ -7182,8 +7182,8 @@ losslessly represents the same reference image.</li>
   <span class="tt">image/png</span> Internet Media type, under the
   <span class="tt">image</span> top level type.
   This appendix is in conformance with
-    <a href="http://www.ietf.org/rfc/rfc4288.txt">BCP 13</a> and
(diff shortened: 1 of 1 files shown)


svgeesus:
(will rebase before merging)

svgeesus:
> There is an updated version of this document, in https://www.w3.org/2020/01/registering-mediatypes.html

I was confused because that is the documentation i was following. But aha, I copy-pasted the annex from another W3C spec and it has an older link. Thanks for noticing!

ProgramMax:
LGTM.
This change would be more than editorial. We should keep it separate from editorial-only changes, which we can easily land in a 2nd Edition update.

Do you think it would be best to wait to land this? Or perhaps land it in a separate branch?

ProgramMax:
Oh nevermind. It is merged already.
We'll cherrypick 2nd Edition editorial updates.

PR #37, opened 2021-11-27
first draft of updated chunk & keyword registry

svgeesus:


lrosenthol:
Could we do this document as simple markdown instead of HTML?  Would be easier to read/edit/maintain.

PR #62, opened 2021-12-17
added improved definitions for Matrix Coefficients and Full Range Flag uses cases

digitaltvguy:
added improved definitions for Matrix Coefficients and Full Range Flag uses cases


index.html
```diff
@@ -4206,18 +4206,16 @@ <h4><a name="11cICP">11.3.3.6 <span class="chunk">cICP</span>
...
-    which are used to describe
-    conversion of YCbCr or ICtCp image samples to an RGB color space.
-    However PNG image samples are explicitly RGB,
-    so the matrix coefficients are not defined in the cICP chunk.
-    Decoders should always assume a matrix coefficient value of 0
-    ("Identity") from the source.</p>
+  <p>Note that in Coding-Independent Code Points
+    <a href="#2-ITU-T-H.273"><span class="NormRef">[ITU-T H.273]</span></a>, the 
+    third video coding point defines matrix coefficients which are used to describe
+    conversion from other color representations to RGB. PNG image samples explicitly RGB,
```
svgeesus: The first part of this paragraph is great but the second is a bit wooly. 
Instead of saying that there is a default, say that it will be 0 for RGB data.
Don't say that decoders should assume 0 (that is leftover text for when the field was missing).

index.html
```diff
@@ -4206,18 +4206,16 @@ <h4><a name="11cICP">11.3.3.6 <span class="chunk">cICP</span>
...
+    so the matrix coefficients default to a value of 0 which is defined as "RGB Identity".
+    Decoders should always assume a matrix coefficient value of 0 until the PNG specification 
+    allows other color representations that require conversion to RGB.</p>
```
svgeesus: I would prefer to say that this field allows for forwards compatibility, if PNG adds other color representations.
The current text sounds like a promise to add them.

index.html
svgeesus: suggest adding (YCbCr, ICtCp) after "other color representations"

index.html
```diff
@@ -4239,23 +4237,19 @@ <h4><a name="11cICP">11.3.3.6 <span class="chunk">cICP</span>
...
 <td class="Regular">Matrix Coefficients</td>
 <td class="Regular">Describes the matrix coefficients used in deriving luma and chroma 
 signals from the green, blue and red as defined in 
 <a href="#2-ITU-T-H.273"><span class="NormRef">[ITU-T H.273]</span></a><a</td>. 
-Currently, PNG is "RGB ONLY" and therefore a value of "1" is the default and indicates 
-"RGB Identity" for the "MATCOEFFS" chunk. Even though Matrix Coefficients aren't needed 
-for RGB sources, we are preserving the full CICP specification for future use-cases where 
-other color representations, that are typical in video imagery might be used 
-(i.e.- YCbCr, ICtCp).
+This value defaults to "0" or "RGB Identity", since PNG is explicitly RGB.
```
svgeesus, line 4245: I would say the value is always 0 for RGB data.
"Default" is imprecise. Does it mean if I specify some other value it gets treated as zero?

index.html
svgeesus: Oh and in HTML  that should be

```html
YC<sub>b</sub>C<sub>r</sub>, IC<sub>T</sub>C<sub>P</sub>
```

digitaltvguy: Fixed with: YC<sub>B</sub>C<sub>R/sub>, IC<sub>T</sub>C<sub>P</sub>.

Capitalized sub-script "B", and "R"

index.html
digitaltvguy: Addressed this imprecise text between matrix coefficients and full-range flag (which is combined in H.273 and very imprecise as well).

index.html
digitaltvguy: Fixed with this:

  <p>Note: In Coding-Independent Code Points <a href="#2-ITU-T-H.273"><span class="NormRef">[ITU-T H.273]</span></a>, the 
    third video coding point defines matrix coefficients which are used to describe
    conversion from other color representations to RGB. PNG image samples are explicitly RGB,
    so the matrix coefficients must default to a value of 0 which is labeled in 
    conversions from other colour representations to RGB. PNG image samples are explicitly RGB,
    so the matrix coefficients value must be set to "0" which is labeled in 
    <a href="#2-ITU-R-BT.709"><span class="NormRef">[ITU-R BT.709]</span></a> as "RGB Identity".
    Decoders must always assume a matrix coefficient value of 0 until the PNG specification 
    allows color representations other than RGB.</p>
    Decoders will always assume a matrix coefficient value of 0 but this value will remain in PNG as per 
    <a href="#2-ITU-T-H.273"><span class="NormRef">[ITU-T H.273]</span></a> for forward compatibility.</p>

index.html
digitaltvguy: How is this:

  <p>Note: In Coding-Independent Code Points <a href="#2-ITU-T-H.273"><span class="NormRef">[ITU-T H.273]</span></a>, the 
    third video coding point defines matrix coefficients which are used to describe
    conversion from other color representations to RGB. PNG image samples are explicitly RGB,
    so the matrix coefficients must default to a value of 0 which is labeled in 
    conversions from other colour representations to RGB. PNG image samples are explicitly RGB,
    so the matrix coefficients value must be set to "0" which is labeled in 
    <a href="#2-ITU-R-BT.709"><span class="NormRef">[ITU-R BT.709]</span></a> as "RGB Identity".
    Decoders must always assume a matrix coefficient value of 0 until the PNG specification 
    allows color representations other than RGB.</p>
    Decoders will always assume a matrix coefficient value of 0 but this value will remain in PNG as per 
    <a href="#2-ITU-T-H.273"><span class="NormRef">[ITU-T H.273]</span></a> for forward compatibility.</p>

digitaltvguy:
How do I resolve each note correctly?  Not sure if I did.

Thank you,
Chris


From: Chris Lilley ***@***.***>
Date: Saturday, December 18, 2021 at 8:55 AM
To: w3c/PNG-spec ***@***.***>
Cc: Seeger, Chris (NBCUniversal) ***@***.***>, State change ***@***.***>
Subject: [EXTERNAL] Re: [w3c/PNG-spec] added improved definitions for Matrix Coefficients and Full Range Flag uses cases (PR #62)

@svgeesus commented on this pull request.

________________________________

In index.html<https://urldefense.com/v3/__https:/github.com/w3c/PNG-spec/pull/62*discussion_r771827893__;Iw!!PIZeeW5wscynRQ!7I1ZyrNMD3lxdVuLaIgZb1Xlz0xnvc7GU4IaBwXhjlxYqy-EWyIzsNs7hfNudZEPHw$>:

>

-  <p>Note that Coding-Independent Code Points

-    <a href="#2-ITU-T-H.273"><span class="NormRef">[ITU-T H.273]</span></a>

-    also define a fourth field, matrix coefficients,

-    which are used to describe

-    conversion of YCbCr or ICtCp image samples to an RGB color space.

-    However PNG image samples are explicitly RGB,

-    so the matrix coefficients are not defined in the cICP chunk.

-    Decoders should always assume a matrix coefficient value of 0

-    ("Identity") from the source.</p>

+  <p>Note that in Coding-Independent Code Points

+    <a href="#2-ITU-T-H.273"><span class="NormRef">[ITU-T H.273]</span></a>, the

+    third video coding point defines matrix coefficients which are used to describe

+    conversion from other color representations to RGB. PNG image samples explicitly RGB,

Oh and in HTML that should be

YC<sub>b</sub>C<sub>r</sub>, IC<sub>T</sub>C<sub>P</sub>

index.html
```diff
@@ -4162,11 +4162,12 @@ <h4><a name="11cICP">11.3.3.6 <span class="chunk">cICP</span>
 <p>The four-byte chunk type field contains the decimal values</p>
 
 <pre>
+  <!--- What are these numbers??? -->
```
svgeesus, line 4154: The decimal values of the bytes corresponding to the type chunk

svgeesus (approved):
This looks good to me

PR #66, opened 2021-12-20
Updated description of four decimal values describing chunks 11.3.3.1-6

digitaltvguy:
Updated description of four decimal values describing chunks 11.3.3.1-6 to make it more readable for those who aren't familiar with the standards syntax.


svgeesus (approved):
Sure, that makes it clearer

ProgramMax:
Agreed. Merging.

PR #67, opened 2021-12-23
removed unnecessary date references from ITU document links, refined definitions, reformated lines

digitaltvguy:
removed unnecessary date references from ITU document links, refined definitions, reformated lines


ProgramMax (approved):
LGTM

Issue #48, opened 2021-12-13
Extensions should link to Guidance for New Chunk Types


svgeesus:
The [Extensions Document](https://w3c.github.io/PNG-spec/extensions/Overview.html) should link to [Guidance for New Chunk Types](https://w3c.github.io/PNG-spec/#B-NewChunksAppendix) to make it more likely that any new chunk registrations follow best practice

PR #72, opened 2022-01-24
Guidance

svgeesus:
Extensions should link to Guidance for New Chunk Types, #48


ProgramMax (approved):
LGTM

There are some trailing whitespaces but I'm not terribly worried about them.

PR #85, opened 2022-01-26
Convert hardcoded refs to SpecRef

svgeesus:
The old References and Bibliography are still there, for checking, but will later be deleted

ISO website is down :) so some missing references at the moment

### DIFF
--- a/index.html
+++ b/index.html
@@ -279,8 +279,8 @@ interchange</i>.</p>
 information exchange between systems &mdash; High-level data link
 control (HDLC) procedures &mdash; Frame structure</i>.</p>
 
-<!-- Maintain a fragment named "2-ISO 8859-1" to preserve incoming links to it -->
-<p class="NormRefDef" id="2-ISO 8859-1">ISO/IEC
+<!-- Maintain a fragment named "2-ISO-8859-1" to preserve incoming links to it -->
+<p class="NormRefDef" id="2-ISO-8859-1">ISO/IEC
 8859-1:1998, <i>Information technology &mdash; 8-bit
 single-byte coded graphic character sets &mdash; Part 1: Latin
 alphabet No. 1</i>.<br class="xhtml" />
@@ -8202,8 +8202,8 @@ Imagery</i>. Springer-Verlag, New York, 1989. ISBN
 <dd>ISO 3664:2000, <i>Viewing conditions &mdash; Graphic
 technology and photography</i>.</dd>
 
-<!-- Maintain a fragment named "G-ITU-R BT.709" to preserve incoming links to it -->
-<dt id="G-ITU-R BT.709">
+<!-- Maintain a fragment named "G-ITU-R-BT.709" to preserve incoming links to it -->
+<dt id="G-ITU-R-BT.709">
 <a name="ITU-R BT.709">[ITU-R BT.709]</a></dt>
 
 <dd>International Telecommunications Union, <i>Basic Parameter
@@ -8314,8 +8314,8 @@ Manual</i>, 2nd edition. Addison-Wesley, Reading, 1990. ISBN
 Video</i>. John Wiley and Sons, Inc., New York, 1996. ISBN
 0-471-12253-X.</dd>
 
-<!-- Maintain a fragment named "G-SMPTE 170M" to preserve incoming links to it -->
-<dt id="G-SMPTE 170M">
+<!-- Maintain a fragment named "G-SMPTE-170M" to preserve incoming links to it -->
+<dt id="G-SMPTE-170M">
 <a name="SMPTE 170M">[SMPTE 170M]</a></dt>
 
 <dd>Society of Motion Picture and Television Engineers,
@@ -8330,8 +8330,8 @@ for Studio Applications</i>, SMPTE 170M, 1994.</dd>
 mapping and the printing of digital images", <i>ACM Transactions on
 Graphics</i>, vol. 7, no. 3, pp. 249-292, 1988.</dd>
 
-<!-- Maintain a fragment named "G-TIFF 6.0" to preserve incoming links to it -->
-<dt id="G-TIFF 6.0">
+<!-- Maintain a fragment named "G-TIFF-6.0" to preserve incoming links to it -->
+<dt id="G-TIFF-6.0">
 <a name="TIFF 6.0">[TIFF 6.0]</a></dt>
 
 <dd>TIFF<sup>TM</sup> Revision 6.0, Aldus Corporation, June
--- a/index.html
+++ b/index.html
@@ -67,15 +67,78 @@
 	  wgPublicList: "public-png",
 	  wgPatentURI: "https://www.w3.org/groups/wg/png/ipr",
     localBiblio: {
-    "CIPA DC-008": {
-      title: "Exchangeable image file format for digital still cameras: Exif Version 2.32",
-      href: "https://www.cipa.jp/std/documents/download_e.html?DC-008-Translation-2019-E",
-      date: "2019-05-17",
-      publisher: "CIPA",
+      "CIPA DC-008": {
+        title: "Exchangeable image file format for digital still cameras: Exif Version 2.32",
+        href: "https://www.cipa.jp/std/documents/download_e.html?DC-008-Translation-2019-E",
+        date: "2019-05-17",
+        publisher: "Camera & Imaging Products Association",
+      },
+      "ICC-2": {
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -67,15 +67,78 @@
...
+        title: "ISO/IEC 3309:1993, Information Technology — Telecommunications and information exchange between systems — High-level data link control (HDLC) procedures — Frame structure.",
+        date: "1993",
+        publisher: "ISO"
+      },
+      "ISO 8859-1": {
+        title: "ISO/IEC 8859-1:1998, Information technology — 8-bit single-byte coded graphic character sets — Part 1: Latin alphabet No. 1.",
+        date: "1998"
+      },
+      "ISO 15076-1": {
+        title: "ISO Website is down, lol"
```
ProgramMax, line 92: lol

index.html
```diff
@@ -210,8 +273,8 @@ <h2 id="2NormRefs">Normative references</h2>
 information exchange between systems &mdash; High-level data link
 control (HDLC) procedures &mdash; Frame structure</i>.</p>
 
-<!-- Maintain a fragment named "2-ISO-8859-1" to preserve incoming links to it -->
-<p class="NormRefDef" id="2-ISO-8859-1">ISO/IEC
+<!-- Maintain a fragment named "2-ISO 8859-1" to preserve incoming links to it -->
+<p class="NormRefDef" id="2-ISO 8859-1">ISO/IEC
```
ProgramMax, line 277: The old fragment had the dash in it. We need to preserve it.

index.html
```diff
@@ -3859,14 +3910,14 @@ <h2><a name="srgb-standard-colour-space"><span class="chunk">sRGB</span>
 
 <p>When the <span class="chunk">sRGB</span> chunk is present, it
 is recommended that decoders that recognize it and are capable of
-colour management <a href="#ICC"><span class=
-"bibref">[ICC]</span></a> ignore the <a href="#gama-image-gamma"><span
+colour management
+ignore the <a href="#gama-image-gamma"><span
```
ProgramMax, line 3914: Just confirming this removal was intentional.
Maybe it was removed because the ISO site is down?

index.html
```diff
@@ -3859,14 +3910,14 @@ <h2><a name="srgb-standard-colour-space"><span class="chunk">sRGB</span>
...
 class="chunk">gAMA</span></a> and <a href="#chrm-primary-chromaticities-and-white-point"><span class=
 "chunk">cHRM</span></a> chunks and use the <span class=
 "chunk">sRGB</span> chunk instead. Decoders that recognize the
 <span class="chunk">sRGB</span> chunk but are not capable of
-colour management <a href="#ICC"><span class=
-"bibref">[ICC]</span></a> are recommended to ignore the <a href=
+colour management
+are recommended to ignore the <a href=
```
ProgramMax, line 3920: (Same here)

index.html
```diff
@@ -4942,14 +4980,11 @@ <h2>Encoder gamma
 "xref">Gamma and chromaticity</span></a> for a brief introduction
 to gamma issues.</p>
 
-<p>PNG encoders capable of full colour management <a href=
-"#ICC"><span class="bibref">[ICC]</span></a> will perform more
+<p>PNG encoders capable of full colour management will perform more
```
ProgramMax, line 4983: (Same here)

index.html
```diff
@@ -5138,14 +5173,11 @@ <h2>Encoder colour
 "xref">Gamma and chromaticity</span></a> for references to colour
 issues.</p>
 
-<p>PNG encoders capable of full colour management <a href=
-"#ICC"><span class="bibref">[ICC]</span></a> will perform more
+<p>PNG encoders capable of full colour management will perform more
```
ProgramMax, line 5176: (Same here)

index.html
```diff
@@ -6414,8 +6439,7 @@ <h2>Decoder gamma
 "xref">Gamma and chromaticity</span></a> for a brief introduction
 to gamma issues.</p>
 
-<p>Viewers capable of full colour management <a href=
-"#ICC"><span class="bibref">[ICC]</span></a> will perform more
+<p>Viewers capable of full colour management  will perform more
```
ProgramMax, line 6442: (Same here)

index.html
```diff
@@ -8293,13 +8308,13 @@ <h2 class="Annex" id="G-References">
 Video</i>. John Wiley and Sons, Inc., New York, 1996. ISBN
 0-471-12253-X.</dd>
 
-<!-- Maintain a fragment named "G-SMPTE-170M" to preserve incoming links to it -->
-<dt id="G-SMPTE-170M">
-<a name="SMPTE-170M">[SMPTE-170M]</a></dt>
+<!-- Maintain a fragment named "G-SMPTE 170M" to preserve incoming links to it -->
+<dt id="G-SMPTE 170M">
+<a name="SMPTE 170M">[SMPTE 170M]</a></dt>
```
ProgramMax, line 8313: The old fragment had a dash and needs to be preserved.
The new fragment (<a name="...") can have it removed if spaces are allowed in fragments. I assume they are?

ProgramMax (changes requested):

index.html
svgeesus: Yeah, will add details once I have them

index.html
svgeesus: Oops my bad, a search-and-replace that clobbered something it shouldn't

index.html
svgeesus: No, it is because that is a rather spurious link for "colour management exists". 
It is to the bibliography, and just links to color.org - it is separate from the normative [[icc]] reference which is to ICC Profile 4.3.

index.html
svgeesus: That whole section will go away because it is replaced by the automatic ReSpec generated references.

I'm unsure how to preserve links to old bibliographic entries (apart from hand-editing the generated references to add them back in)

index.html
ProgramMax: Okay. Makes sense to me. (It was rather odd.)
So these are indeed intentional, right?

svgeesus: Yes.

index.html
svgeesus: Details for these two now added

index.html
ProgramMax: I think we still want this id to have a dash, right?

index.html
```diff
@@ -8181,9 +8202,9 @@ <h2 class="Annex" id="G-References">
 <dd>ISO 3664:2000, <i>Viewing conditions &mdash; Graphic
 technology and photography</i>.</dd>
 
-<!-- Maintain a fragment named "G-ITU-R-BT709" to preserve incoming links to it -->
-<dt id="G-ITU-R-BT709">
-<a name="ITU-R-BT709">[ITU-R-BT709]</a></dt>
+<!-- Maintain a fragment named "G-ITU-R BT.709" to preserve incoming links to it -->
```
ProgramMax, line 8205: I think we need to preserve the dash.

index.html
ProgramMax: Let's preserve it for now. If we end up deleting it as part of the move to ReSpec, that is okay. But let's break it there, not here.

index.html
```diff
@@ -8309,9 +8330,9 @@ <h2 class="Annex" id="G-References">
 mapping and the printing of digital images", <i>ACM Transactions on
 Graphics</i>, vol. 7, no. 3, pp. 249-292, 1988.</dd>
 
-<!-- Maintain a fragment named "G-TIFF-6.0" to preserve incoming links to it -->
-<dt id="G-TIFF-6.0">
-<a name="TIFF-6.0">[TIFF-6.0]</a></dt>
+<!-- Maintain a fragment named "G-TIFF 6.0" to preserve incoming links to it -->
```
ProgramMax, line 8333: (Same here)

ProgramMax (changes requested):
LGTM except let's preserve the dashes for now.

index.html
svgeesus: It is in the section that will be deleted - the old, manual, normative references section.

I can see someone linking to a particular spec section, (like a chunk definition) or to a defined term, but doubt there are many/any links to an entry in the references. They would just link directly to that reference, surely.

index.html
svgeesus: This *is* the move to ReSpec :)

svgeesus:
Okay, dashes preserved

index.html
svgeesus: But okay

ProgramMax (approved):
LGTM

PR #91, opened 2022-02-08
Add missing mention of CICP, fix #83

### LINKED ISSUE #83: Update Colour Spaces section to mention CICP
The spec [currently says](https://w3c.github.io/PNG-spec/#4Concepts.ColourSpaces):

```html
<p>The RGB colour space in which colour samples are situated may
be specified in one of three ways:</p>
```

and then lists

1. ICC profile
2. sRGB chunk
3. gamma and chromaticity

It needs to be updated to account for the new fourth method, CICP.

svgeesus:


ProgramMax (approved):
LGTM

PR #109, opened 2022-04-05
Use correct ReSpec syntax for linking to terms.

svgeesus:
It helps if I actually [read the documentation](https://respec.org/docs/#definitions-and-linking) on ReSpec linking to defined terms. I was using `[[defined term]]` but it should be `[=defined term=]`

These are just fixing  the recently-introduced APNG-related definitions.

The other unused definitions which ReSpec complains about will need work to keep the old and the new IDs (perhaps like this:

```html
<dfn>
  <a id="3oldId"/>
  <dt>Term with no ID</dt>
</dfn>
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -485,7 +485,7 @@ necessarily the best possible image.
   <dd>Optional animation, consisting of a series of frames.
     The first frame may be,
     but need not be,
-    the [[static image]].
+    the [=static image=].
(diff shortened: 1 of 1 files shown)


ProgramMax (approved):
LGTM.

I missed it, too. :) Good catch.

PR #110, opened 2022-04-05
Remove "Relationship to W3C PNG", add up to date changes section

svgeesus:
~~Remove remaining "this International Standard", replacing with "this specification".~~

Remove old, confusingly named "Relationship to W3C PNG" section, which was about first to second edition changes and [still exists in PNG Second Edition](https://www.w3.org/TR/2003/REC-PNG-20031110/index.html#F-Relationship) and replace with modern Changes section.

The previous editor and author names, which were also in that old section, are already in the header of the document.

### DIFF
--- a/index.html
+++ b/index.html
@@ -7976,10 +7976,10 @@ specified in this specification.</li>
 <li>No chunks appear in the PNG datastream other than those
 specified in this specification or those defined
 according to the rules for creating new chunk types as defined in
-this specification.</li>
+this International Standard.</li>
 
 <li>The PNG datastream is encoded according to the rules of this
-specification.</li>
+International Standard.</li>
 </ol>
 </section>
 
@@ -7990,7 +7990,7 @@ specification.</li>
 <h2>Conformance of PNG
 encoders</h2>
 
-<p>A PNG encoder conforms to this specification if it
+<p>A PNG encoder conforms to this International Standard if it
 satisfies the following conditions.</p>
 
 <!-- <ol start="1"> --><ol>
@@ -8014,12 +8014,12 @@ fields.</li>
 <h2>Conformance of PNG
 decoders</h2>
 
-<p>A PNG decoder conforms to this specification if it
+<p>A PNG decoder conforms to this International Standard if it
 satisfies the following conditions.</p>
 
 <!-- <ol start="1"> --><ol>
 <li>It is able to read any PNG datastream that conforms to this
-specification, including both public and private chunks
+International Standard, including both public and private chunks
 whose types may not be recognized.</li>
 
 <li>It supports all the standardized critical chunks, and all the
@@ -8081,7 +8081,7 @@ ordering rules.</li>
 <h2>Conformance of PNG
 editors</h2>
 
-<p>A PNG editor conforms to this specification if it satisfies the following conditions.</p>
+<p>A PNG editor conforms to this International Standard if it satisfies the following conditions.</p>
 
 <ol>
 <li>It conforms to the requirements for PNG encoders.</li>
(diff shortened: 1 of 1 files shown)


palemieux (changes requested):
_International Standard_ will be required by ISO

svgeesus:
For [specifications submitted as PAS](https://www.w3.org/2010/04/pasfaq) there is no requirement to use ISO-ese

palemieux:
> For [specifications submitted as PAS](https://www.w3.org/2010/04/pasfaq) there is no requirement to use ISO-ese

PAS must be turned into International Standards within 6 years... so (a) I would format the document with the assumption that it becomes an IS and (b) question why go through PAS at all?

svgeesus:
@palemieux please compare the text of 

Mathematical Markup Language (MathML) Version 3.0 2nd Edition
W3C Recommendation 10 April 2014 
https://www.w3.org/TR/2014/REC-MathML3-20140410/

with the text of

ISO/IEC 40314:2016
Information technology — Mathematical Markup Language (MathML) Version 3.0 2nd Edition
https://www.iso.org/standard/58439.html

in particular [2.1.1 General Considerations ](https://www.w3.org/TR/2014/REC-MathML3-20140410/chapter2.html#fund.xmlgeneral) where we find the phrase:

>  These rules are defined by this specification

in both the W3C and the ISO versions.

MathML 3.0 went through the PAS process.

palemieux:
@svgeesus ISO editors have significantly tightened the rules since 2016 (coincidentally). Why bother to make the change? Why does keeping "International Standard" hurt if the goal is for the spec to become one ultimately? Do you think ISO will refuse to publish it if the term "International Standard" is present.

svgeesus:
> ISO editors have significantly tightened the rules since 2016 (coincidentally). 

Good to know. Got a pointer to the rule changes?

> Why bother to make the change? Why does keeping "International Standard" hurt if the goal is for the spec to become one ultimately? 

Because

1. This here is an International Standard
2. We plan that eventually, this will be an International Standard
3. This is derived from a spec that was an International Standard but none of the changes have been approved by ISO and we haven't even published a FPWD

are not equivalent statements. "This International Standard" was appropriate (because true) in PNG Second Edition.

> Do you think ISO will refuse to publish it if the term "International Standard" is present.

That isn't my particular concern, no. More about not misleading the readership regarding the current status.

palemieux:
> More about not misleading the readership regarding the current status.

Ok. I see it the other way: we should keep "International Standard" since the intent is to create a new edition of the existing "International Standard", not create a separate spec. The parallel in my mind is that a draft ISO specification says "International Standard" before it is approved (the cover page says "DRAFT" of course).

palemieux:
> Do you have a pointer to the post-2016 rule changes?

It is not a change of rule, but how strongly they are enforced.

My basic point remains: if this is a new edition, then we should retain "International Standard", which captures the intent.

ProgramMax:
FWIW, I do think the large "Draft" is sufficient to explain that it is not yet an international standard.

And given that, I would prefer to have as few changes as possible between the draft and the published version. Ideally, the only change is that large "Draft". :)

I don't have enough conclusive precedent or a clear rule to know what the right answer is. But my personal feeling is leaning toward leaving "international standard" in place. An international airport is called that while it is still being constructed. The plans are in place for it to be one.

index.html
```diff
@@ -8011,7 +8019,7 @@ <h2>Conformance of PNG
 
 <!-- <ol start="1"> --><ol>
 <li>It is able to read any PNG datastream that conforms to this
-International Standard, including both public and private chunks
+specification, including both public and private chunks
```
palemieux, line 8022: @svgeesus This and other changes were not reverted?

palemieux (changes requested):

index.html
svgeesus: Thanks, please check again now

palemieux (approved):

PR #118, opened 2022-05-02
fix up broken structure (missing two closing section tags)

svgeesus:
The table of contents was broken, putting much of the spec into section 5! This was caused by two missing `</section>` which this PR fixes.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2071,6 +2071,7 @@ available for private use.
 </p>
 </section>
 </section>
+</section>
 
 <!-- Maintain a fragment named "5DataRep" to preserve incoming links to it -->
 <section id="5DataRep">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2071,6 +2071,7 @@ <h2>Extensions</h2>
 </p>
 </section>
 </section>
+</section>
```
ProgramMax, line 2074: I went through this carefully and I think this should be on line 2030.
The existing spec already has section 4.8 Error Handling.
The new APNG information was added just before this section. But it wasn't closed completely. As a result, the Error Handling section is currently 5.9.4 (3 levels deep). It should only be 2 levels deep.

If this closing section tag were on line 2030, it would close out the APNG information and keep the Error Handling section 2 levels deep (5.9 Error Handling).

ProgramMax (changes requested):

PR #130, opened 2022-06-14
Remove hard-coded heading numbers

### LINKED ISSUE #106: Remove hard-coded section numbers
A lot of hard-coded section numbers are now incorrect because of the way ReSpec wants to starts its sections.

For example, "See 6.1: Colour types and values". It is now 7.1.

Rather than hard-code the section numbers, this should link to the section name like [[[#section-name]]]. ReSpec will then include the section numbers it generated.

(Many of these hard-coded section numbers can be found just by searching for "see".)

palemieux:
Closes #106

### DIFF
--- a/index.html
+++ b/index.html
@@ -429,7 +429,7 @@ definitions apply.</p>
 
 <dl>
 <!-- Maintain a fragment named "3alpha" to preserve incoming links to it -->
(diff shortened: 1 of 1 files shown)


svgeesus (approved):
LGTM

Issue #122, opened 2022-06-01
Use ReSpec formatting for internal links


ProgramMax:
ReSpec lets you link to other parts of the page via [[[#fragment-name]]]. This will auto-populate with the section number.

We should update internal links to use this formatting.

This is part of #86.

ProgramMax:
Note to self: The definition of "chromaticity" is an odd case because it currently includes the "(CIE)" part.
I'm skipping it for now. But come back to it.

svgeesus:
@ProgramMax said:

>The definition of "chromaticity" is an odd case because it currently includes the "(CIE)" part.

Spec currently says

> chromaticity (CIE)
>    pair of values x,y that precisely specify a colour, except for the brightness information.

suggested rewording

> chromaticity 
>    pair of CIE x,y values [[COLORIMETRY]] that precisely specify a colour, except for the brightness information.

This also adds a reference to the [defining CIE standard](http://cie.co.at/publications/colorimetry-4th-edition)

ProgramMax:
The way I was working on this is to have two views of the same file open.
In one view, I search for 'id="'.
In the other view, I then search for 'href="#foo' and update them all to '<a>foo</a>'.

Then just go down the list of all 'id="' matches until the commit is getting long. I was splitting it up across multiple commits to keep reviews easy. The last match I left off on was 'id="3sampleDepth'.

One thing to be aware of is how '<a>foo</a>' finds matches. It will match '<section><h2>foo</h2>' and '<dfn><dt>foo</dt>'. These can have any ids added to them, such as '<section id="3foo"><h2>foo</h2>'. It also matches pluralization and other heading levels. What this means is you might encounter 'href="#foo"' but need to update it to '<a>bar</a>' depending on what it matches.

You may also come across '[=foo=]', '[[foo]]', and '[[[foo]]]' from previous edits that should be '<a>foo</a>'.

(As a result of this process, #123 is also addressed.)

PR #131, opened 2022-06-15
Regularise `chromaticity` cross references to use ReSpec, add CIE Colorimetry reference

svgeesus:
As [noted](https://github.com/w3c/PNG-spec/issues/122#issuecomment-1154313952) by @ProgramMax the definition for chromaticity was a bit odd in that it added a parenthetical (CIE) which then showed up in all cross references.

I moved the CIE mention to the definition, and added `[[COLORIMETRY]]` which is the defining standard for chromaticity and is already referenced in the definition for luminance.

Now uses `[=chromaticity=]` consistently.

### DIFF
--- a/index.html
+++ b/index.html
@@ -550,9 +550,9 @@ image.</dd>
 
 <!-- Maintain a fragment named "3chromaticity" to preserve incoming links to it -->
 <dfn id="3chromaticity">
-<dt>chromaticity (CIE)</dt></dfn>
+<dt>chromaticity</dt></dfn>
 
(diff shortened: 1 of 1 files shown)


ProgramMax (approved):
LGTM.

I sifted through the history to remember the previous issue we ran into with the chromaticity definition.
It was originally worded was "chromaticity" when linking to the definition. Allowing ReSpec to reformat the links, they were replaced with "chromaticity (CIE)", which was the definition's text.

This commit puts the links back to "chromaticity". So the previous issue is actually solved better here.

Thank you.

PR #142, opened 2022-07-19
 Update figures, fix #90

### LINKED ISSUE #90: Update figures
Many figures are fine right now.
But some either look dated or have clipped text. For example, figure 7:
https://w3c.github.io/PNG-spec/#encoding-png-image

The "Scanline serialization" text clips outside its box. So does the "Compression" text.
Plus, the squares that represent pixels have a shadow that matches '90s GUIs. (And they could probably use better color selection. Be sure to consider color blindness here.)

This isn't strictly necessary. But it would go a long way toward keeping PNG modern and not feeling like an abandoned format.

svgeesus:
Fixed some issues:

- Hard-coded image widths in pixels
- No webfont (assumed Verdana)
- Required unit on `font-size` missing, so ignored and 16px used
- Ugly drop shadows
- Ugly cyan fill for no reason
- pointless DTDs etc
- stroking of text, yuk

I created a WOFF2 of a ASCII-only subset of Open Sans Medium Condensed. This adds little to the spec size and is shared by all the SVGs, giving a consistent look; condensed font allows the text size to be bigger for readability.

No more text spilling out of boxed.

I also added the `cICP` chunk to the chunks diagrams.


svgeesus:
Oh, I should add [eXIf](https://w3c.github.io/PNG-spec/#eXIf) to the chunk ordering diagrams!

ProgramMax (approved):
LGTM. I won't merge until you've added the eXIf chunk.

Text still spills a bit in chunk-parts.svg.

If we're no longer using the rendered pngs, maybe those should be renamed back to the original file names? (That should be done in a separate commit.) Actually--now that I think about it--if we're adding new chunks to the diagrams then perhaps those should maintain an old copy, too.

What do you think? Maybe leave these with their descriptive names and restore the old files with their fig15 type names? (Then we don't need to setup any redirect rules on the server.)

svgeesus:
W are no longer using the rendered pngs, they can be deleted. They made sense when SVG was super new and required a plug-in for browsers. They don't, any more.

svgeesus:
I'm not seeing any text spill in chunk-parts.svg

image

ProgramMax:
The 3rd Edition doesn't need the old PNG files but the older editions still reference those images.
Previously, I renamed the files. I think we can just rename them back to what they were and only keep them around for the older editions.

I think you're right about the font not spilling. Here is what I'm seeing:
Screenshot 2022-07-20 123406
But I think that is in the GitHub preview where it might not be able to find the font. I notice what I'm seeing has serifs and your image does not.

svgeesus:
> The 3rd Edition doesn't need the old PNG files but the older editions still reference those images.
Previously, I renamed the files. I think we can just rename them back to what they were and only keep them around for the older editions.

The Second edition images are still there on /TR from the previously published editions. I don't see value in continuing to keep around outdated images, and the PNG are extra effort to update compared to the SVG source files.

svgeesus:
> But I think that is in the GitHub preview where it might not be able to find the font.

SVG included in html `img` tags does not fetch external resources such as style sheets, images, or fonts. Presumably the GitHub preview is using `img` not `object`

ProgramMax:
If the old spec stays in tact then I agree that we should delete the PNG renderings of the SVG files.

Issue #11, opened 2021-01-19
Filter description unclear


svgeesus:
Raised by: Björn Höhrmann
Originally raised: [04 Mar 2011](https://lists.w3.org/Archives/Public/png-group/2011Mar/0000.html)

> The second thing is the discussion of filters. Filters map from source
to destination, but section 9.2 isn't quite phrased this way. It's hard
to properly understand the "pixel versus byte" distinction early on in
the section, and the notation in table 9.1 is hard to understand. I got
confused a couple of times in what order things had to be done, what it
really means to have `Filt(x)` in the reconstruction function and so on.

> One ambiguity for instance is that it's not really clear what `Orig(x)`
is since from the perspective of the scanline for `x` there are two
original values for the other variables, before and after they had their
filter applied. The same goes for `Recon(a)`, is that before or after it
had its scanline filter removed? While I figured out the answer, I did
understand these things incorrectly initially.

> (It would have helped a lot had there been some pseudo-code showing how
you get from the color and depth values to the advance width and how you
resolve the aforementioned ambiguity, or at the least something about
how you can decode top-down but need to encode bottom-up or with a separate buffer.)

svgeesus:
Yes this could at minimum do with a diagram showing two or three scanlines in 

1. the original image
2. the filtered image
3. the reconstructed image.

This would help to graphically illustrate the somewhat convoluted

> the byte corresponding to x in the pixel immediately before the pixel containing x

ProgramMax:
This seems to have multiple parts. One of the parts is the interaction between scanlines and filtering.

There are two really great figures to help show that interaction:
https://w3c.github.io/PNG-spec/#encoding-png-image
https://w3c.github.io/PNG-spec/#serializing-and-filtering-scanline

Note that these are both in section 4. The section on filter (9) doesn't reference back to these figures:
https://w3c.github.io/PNG-spec/#9Filters

It would be easy to jump to the filtering section miss these figures. Or even if the doc is read top-to-bottom, by the time someone gets to section 9 they may not think to jump back to section 4.

I think one way to improve understanding on filtering is to simply reference these figures from section 9.

ProgramMax:
There is also a sentence "[Figure 15](https://w3c.github.io/PNG-spec/#paethpredictor-function) shows the relative positions of the bytes x, a, b, and c." with the context of filters in general and not specifically the Paeth filter.

[...]

PR #148, opened 2022-07-25
Add new filter bytes diagram, improve table, see #11

svgeesus:


index.html
```diff
@@ -2937,30 +2944,41 @@ <h2>Filter types for filter method
...
 <tr>
-<td class="Regular"><tt>c</tt> </td>
+<td class="Regular">c </td>
 <td class="Regular">the byte corresponding to b in the pixel immediately before the pixel containing b (or the byte immediately before b, when the bit depth is less than 8).</td>
 </tr>
 </table>
 
-<p><a href="#paethpredictor-function"></a> shows the relative positions of the bytes <tt>x</tt>,
+<figure id="integer-representation-in-png">
+  <object id="figure72" data="figures/filter-bytes.svg" type="image/svg+xml" style="width: 12vw">
```
ProgramMax, line 2976: Since this is a new figure, can we go with an id that isn't related to the section? Perhaps a descriptive name.

ProgramMax (commented):
LGTM but can we change the figure id?

index.html
svgeesus: Sure! It wasn't clear to me how the figure id and the object id were different.

svgeesus: Oh I see, sorry that was a dumb copy-paste error to get the markup, from another figure!

ProgramMax: Oh sorry. I do like the figure id more now. But I actually meant the object id. Sorry about that.

What I meant is "figure72" sounds like it assumes what section it is in.

I don't think the object needs an id at all. The figure does so we can link to it with ReSpec magic. I suspect elsewhere the figure and object both have ids so we can preserve an old fragment identifier while also giving it a new-style name.

I think in this case we can just drop the object id.

svgeesus:
@ProgramMax said:

> LGTM but can we change the figure id?

Done with https://github.com/w3c/PNG-spec/pull/148/commits/15ff9efda2e259da873ff554557a29c084f7707d ok now?

ProgramMax (approved):
No no, that commit updated the figure's id, not the object's id.
(That said, this isn't critical and can be fixed later or ignored if we don't care enough. I'll go ahead and approve.)

Issue #135, opened 2022-06-23
I can do that.


palemieux:
I can do that.
Would you be okay if I do that in a separate pull request, though?

_Originally posted by @ProgramMax in https://github.com/w3c/PNG-spec/pull/128#discussion_r905389847_

ProgramMax:
To add context here, the class named "Definition" was from a style sheet that was added to one edition of the spec and removed in a subsequent revision to that edition. There is no longer a matching class named "Definition".

ProgramMax:
Also `<span class="chunk">`

palemieux:
> Also <span class="chunk">

I think this is a separate issue since we might actually want to style reference chunks differently.

I am not too excited by `<code><a>cICP</a></code>`... maybe there is something better that both (a) conveys visually that it is a chunk name and (b) links to the chunk definition?

ProgramMax:
I agree with you. Even if we don't plan to use it now, it provides an easy access point into styling chunks.

I think there are several others of these to consider, still. For example, `<table class="Regular">` and `<td class="Regular">`?

ProgramMax:
There is also `<figcaption class="Figuretitle">`, `<span class="figref">`, and `<span class="tableref">`...

ProgramMax:
And `<br class="xhtml" />` :D

svgeesus:
> Also `<span class="chunk">`

I think we should keep that one tbh

palemieux:
> > Also `<span class="chunk">`
> 
> I think we should keep that one tbh

We could do `<a class="chunk">cICP</a>`?

ProgramMax:
Note that in the spec there is an example comparing CHNK and cHNk as theoretical, example chunks. These are marked as '<span class="chunk">' and cannot be marked as '<a class="chunk">'. I left these as-is.

PR #154, opened 2022-08-05
Fix internal and definition links

### LINKED ISSUE #122: Use ReSpec formatting for internal links

### LINKED ISSUE #123: Use ReSpec formatting for links to definitions
ReSpec allows using [=term=] to link to defined terms. This will not include section numbers.

Even though it is currently the same as the existing anchor tags, using ReSpec's auto-formatting is forward compatible in case the behavior changes later. We should use it.

palemieux:
Closes #123 
Closes #122

### DIFF
--- a/index.html
+++ b/index.html
@@ -348,8 +348,7 @@ definitions apply.</p>
 <!-- Maintain a fragment named "3alpha" to preserve incoming links to it -->
 <dt><dfn id="3alpha">alpha</dfn></dt>
 
-<dd>a value representing a <span class=
-"Definition"><a>pixel</a></span> degree of opacity. The more
+<dd>a value representing a <a>pixel</a> degree of opacity. The more
 opaque a pixel, the more it hides the background against which
 the image is presented. Zero alpha represents a completely
 transparent pixel, maximum alpha represents a completely opaque
@@ -358,46 +357,34 @@ pixel.</dd>
 <dt><dfn id="3alphaCompaction">alpha compaction</dfn></dt>
 
 <dd>an implicit representation of transparent
-<span class="Definition"><a>pixels</a></span>. If every
-pixel with a specific colour or <span
-class="Definition"><a>greyscale</a></span> value is fully
+<a>pixels</a>. If every
+pixel with a specific colour or <a>greyscale</a> value is fully
 transparent and all other pixels are fully opaque, the
-<span class="Definition"><a>alpha</a></span>
-<span class="Definition"><a>channel</a></span> may be
+<a>alpha</a>
+<a>channel</a> may be
 represented implicitly.</dd>
 
 <!-- Maintain a fragment named "3alphaSeparation" to preserve incoming links to it -->
 <dt><dfn id="3alphaSeparation">alpha separation</dfn></dt>
 
-<dd>separating an <span class=
-"Definition"><a>alpha</a></span> <span class=
-"Definition"><a>channel</a></span> in which every
-<span class="Definition"><a>pixel</a></span> is fully
+<dd>separating an <a>alpha</a> <a>channel</a> in which every
+<a>pixel</a> is fully
 opaque; all alpha values are the maximum value.
 The fact that all pixels are fully opaque is represented implicitly.
 </dd>
 
 <!-- Maintain a fragment named "3alphaTable" to preserve incoming links to it -->
-<dfn id="3alphaTable">
-<dt>alpha table</dt></dfn>
+<dt id="3alphaTable"><dfn>alpha table</dfn></dt>
 
-<dd>indexed table of <span class=
-"Definition"><a>alpha</a></span> <span class=
-"Definition"><a>sample</a></span> values, which in an <span class=
-"Definition"><a>indexed-colour</a></span> image defines the alpha
-sample values of the <span class=
-"Definition"><a>reference image</a></span>. The alpha table has the
-same number of entries as the <span class=
-"Definition"><a>palette</a></span>.</dd>
+<dd>indexed table of <a>alpha</a> <a>sample</a> values, which in an <a>indexed-colour</a> image defines the alpha
+sample values of the <a>reference image</a>. The alpha table has the
+same number of entries as the <a>palette</a>.</dd>
 
 <!-- Maintain a fragment named "3ancillaryChunk" to preserve incoming links to it -->
-<dfn id="3ancillaryChunk">
-<dt>ancillary chunk</dt></dfn>
+<dt id="3ancillaryChunk"><dfn>ancillary chunk</dfn></dt>
 
-<dd>class of <span class=
-"Definition"><a>chunk</a></span></a> that provides additional
-information. A <span class=
-"Definition"><a>PNG decoder</a></span>, without processing an
+<dd>class of <a>chunk</a></a> that provides additional
+information. A <a>PNG decoder</a>, without processing an
 ancillary chunk, can still produce a meaningful image, though not
 necessarily the best possible image.
 <!-- agreed: don't need to define a bit -->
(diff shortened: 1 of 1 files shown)


palemieux:
@ProgramMax @svgeesus There are significant issues with definitions. For example:

* some definitions are not referenced in the body of the documents, e.g., _PNG four-byte signed integer_
* some definitions duplicate normative prose, e.g., _colour type_
* some definitions are inconsistently used/vague, e.g. does the _filter_ definition applies to _filter method_

index.html
```diff
@@ -516,76 +477,62 @@ <h2 id="3Definitions">Definitions</h2>
...
-<dfn id="3deflate">
-<dt>deflate</dt></dfn>
+<dt id="3deflate"><dfn>deflate</dfn></dt>
 
 <dd>name of a particular compression algorithm. This algorithm is
 used, in compression mode 0, in conforming
-<span class="Definition"><a>PNG
-datastreams</a></span>. Deflate is a member of the <a href=
-"#3LZ77"><span class="Definition">LZ77</span></a> family of
+<<a>PNG datastreams</a>. Deflate is a member of the <a>LZ77</a> family of
```
ProgramMax, line 501: There is a double `<<` at the beginning of this line. I don't think that was intentional.

index.html
```diff
@@ -3204,14 +3034,14 @@ <h2>Compression method 0</h2>
...
-given in the deflate specification [[rfc1951]].</p>
+given in the <a>deflate</a> specification [[rfc1951]].</p>
 
 <p>The check value stored at the end of the zlib datastream is
 calculated on the uncompressed data represented by the
 datastream. The algorithm used to calculate this is not the same
 as the CRC calculation used for PNG chunk CRC field values. The
 zlib check value is useful mainly as a cross-check that the
-deflate and inflate algorithms are implemented correctly.
+<a>deflate</a> algorithms are implemented correctly.
```
ProgramMax, line 3044: Here, too.

index.html
```diff
@@ -2978,13 +2977,13 @@ <h2>Compression method 0</h2>
 <p>Only PNG compression method 0 is defined by this International
 Standard. Other values of compression method are reserved for
 future standardization. PNG compression method 0 is
-deflate/inflate compression with a sliding window
+<a>deflate</a> compression with a sliding window
```
ProgramMax, line 2980: This was previously "deflate/inflate".
I _think_ when adding the link, you accidentally removed "/inflate".
I question myself there because it happens elsewhere too so maybe it was intentional?

But I believe it should still include the "inflate" wording in all those cases. I'll point them out.

index.html
```diff
@@ -3258,7 +3257,7 @@ <h2><span class="chunk">IHDR</span> Image
 
 <p>Compression method is a single-byte integer that indicates the
 method used to compress the image data. Only compression method 0
-(deflate/inflate compression with a sliding window of at most
+(<a>deflate</a> compression with a sliding window of at most
```
ProgramMax, line 3260: Here, too.

index.html
```diff
@@ -4373,7 +4372,7 @@ <h2><span class="chunk">zTXt</span>
 <a href="#11tEXt"></a>). The keyword is not
 compressed. The compression method entry defines the compression
 method used. The only value defined in this International
-Standard is 0 (deflate/inflate compression). Other values are
+Standard is 0 (<a>deflate</a> compression). Other values are
```
ProgramMax, line 4375: Here, too.

index.html
```diff
@@ -8325,8 +8324,8 @@ <h2>PNG web site</h2>
 This page is a central location for current information about PNG
 and PNG-related tools.</p>
 
-<p>Additional documentation and portable C code for deflate,
-inflate, and an optimized implementation of the CRC algorithm are
+<p>Additional documentation and portable C code for <a>deflate</a>,
+and an optimized implementation of the CRC algorithm are
```
ProgramMax, line 8328: Here, too.

ProgramMax (approved):
Mostly LGTM.
Two potential changes: A double `<<` and making sure removal of "inflate" was intentional. (Perhaps both deflate and inflate are mentioned in places where both the encoder and decoder are concerned?)

I'll create separate GitHub issues for your other comments (significant issues with definitions). They should be resolved. But some might need discussion.


I see the change from `<dfn><dt>blah</dt></dfn>` to the other nesting order. That change seems correct. `<dl>` changed from "definition list" to "description list" and a definition is now part of a description. Good catch on that!

I also see a lot of `<span class="Definition">` being removed. This is good but if we were to be nitpicky, it should happen in a separate pull request.

I like the `<dfn data-lt="composited|composite">composite (verb)</dfn>`. I didn't know that existed. :)

ProgramMax:
This pull request also contributes toward #135

index.html
svgeesus: Should now be "deflate algorithm is implemented correctly"

svgeesus:
> some definitions are not referenced in the body of the documents, e.g., PNG four-byte signed integer

But fairly close terms are used and are probably what was meant, eg in

> network [byte order](https://w3c.github.io/PNG-spec/#3byteOrder)
    byte order in which the most significant byte comes first, then the less significant bytes in descending order of significance ([MSB](https://w3c.github.io/PNG-spec/#3MSB) [LSB](https://w3c.github.io/PNG-spec/#3LSB) for two-byte integers, [MSB](https://w3c.github.io/PNG-spec/#3MSB) B2 B1 [LSB](https://w3c.github.io/PNG-spec/#3LSB) for four-byte integers).

I suspect that some terms got reworded over time without checking that the defined term and the usage stayed in sync.

index.html
```diff
@@ -358,53 +357,40 @@ <h2 id="3Definitions">Definitions</h2>
 <dt><dfn id="3alphaCompaction">alpha compaction</dfn></dt>
 
 <dd>an implicit representation of transparent
-<span class="Definition"><a>pixels</a></span>. If every
-pixel with a specific colour or <span
-class="Definition"><a>greyscale</a></span> value is fully
+<a>pixels</a>. If every
```
svgeesus, line 360: Does respec deal with singular and plural? (I know bikeshed does, or tries to, not always successfully)? Here `pixel` and `pixels` should link to the same definition.

index.html
```diff
@@ -358,53 +357,40 @@ <h2 id="3Definitions">Definitions</h2>
...
+<dt id="3alphaTable"><dfn>alpha table</dfn></dt>
+
+<dd>indexed table of <a>alpha</a> <a>sample</a> values, which in an <a>indexed-colour</a> image defines the alpha
+sample values of the <a>reference image</a>. The alpha table has the
+same number of entries as the <a>palette</a>.</dd>
 
 <!-- Maintain a fragment named "3ancillaryChunk" to preserve incoming links to it -->
-<dfn id="3ancillaryChunk">
-<dt>ancillary chunk</dt></dfn>
+<dt id="3ancillaryChunk"><dfn>ancillary chunk</dfn></dt>
```
svgeesus, line 384: Yeah it always struck me as odd to have 

```html
<dfn><dt>term</dt></dfn>
<dd>definition</dd>
```

index.html
```diff
@@ -437,77 +417,58 @@ <h2 id="3Definitions">Definitions</h2>
...
-<span class=
-"Definition"><a>ancillary chunk</a></span>.
+<a>critical chunk</a> or an
+<a>ancillary chunk</a>.
 </dd>
 
 <!-- Maintain a fragment named "3colourType" to preserve incoming links to it -->
-<dfn id="3colourType">
-<dt>colour type</dt></dfn>
+<dt id="3colourType"><dfn>colour type</dfn></dt>
```
svgeesus, line 461: In the text, `PNG image type ` seems to be used instead of `colour type`, and it is a better term because alpha is not a color.

svgeesus:
Another example is `delivered image` which is defined by according to respec "not used" i.e. no marked up in the body text as a use of a defined term.

It is used (and defined) though, and that usage is better than the one in the list of definitions:

> The delivered image is constructed from the PNG image obtained by decoding a PNG datastream. No specific format is specified for the delivered image. A viewer presents an image to the user as close to the appearance of the original source image as it can achieve.

We had agreed to keep, in the list of definitions, only those terms which are used but not defined in the body of the specification. Things like network byte order, or indeed byte.  So here that should become

> The `<dfn>`delivered image`</dfn>` is constructed from the `<a>`PNG image`</a>` obtained by decoding a `<a>`PNG datastream`</a>`. No specific format is specified for the delivered image. A viewer presents an image to the user as close to the appearance of the original source image as it can achieve.

svgeesus:
`lossless compression` and `lossy compression` are defined but not used. However the term `lossless` or `loslessly` are used a bunch of times. I would fix this by keeping the term in the definitions list, but putting the `dfn` tags around "lossless" and adding the `data-lt` to add "losslessly"

svgeesus:
in **14.2 Behaviour of PNG editors** we have

> A "PNG editor" is defined as a program that reads a PNG datastream ...

I suggest that this should be made the canonical definition, and the "PNG Editor" removed from the list of defined terms.

svgeesus:
I recall that a definition of one of the PNG integer types was added specifically in response to a reported problem with the spec and exactly what the range f allowable numbers was.

Now that we have `dt` and `dfn` in the correct order, I suggest we keep them in the definitions list but marked up as

```html
<dt>PNG <dfn>four-byte unsigned integer</dfn>
<dd> ...</dd>
```

because we have plenty of instances of "four-byte unsigned integer" that could become `<a>`four-byte unsigned integer`</a>` and this clarifies that while there are several ways to represent that type, this is the way PNG consistently uses.

index.html
ProgramMax: The docs say it does.

svgeesus: Cool, great!

palemieux:
@svgeesus @ProgramMax I recommend we merge this as-is, and I can start fixing the dangling definitions per the above. Specifically:

_We had agreed to keep, in the list of definitions, only those terms which are used but not defined in the body of the specification. Things like network byte order, or indeed byte._ https://github.com/w3c/PNG-spec/pull/154#issuecomment-1212961209

ProgramMax:
I agree. :)

PR #159, opened 2022-08-10
Norm ref RFC1952 for CRC algorithm.  Fix #12

### LINKED ISSUE #12: CRC calculation not well defined
Raised by: Björn Höhrmann
Originally raised: [04 Mar 2011](https://lists.w3.org/Archives/Public/png-group/2011Mar/0000.html)

> The third issue is the CRC calculation. These days there are very many
things that go by "CRC" with various polynominals and whatnot. Since I'd
rather not code my own implementation, or use the specification's code,
I went looking for the right library, but that's not very easy. It would
have helped if there had been a reference, like saying this is the same
CRC that is used by gzip.

> Similarily, in 5.1 "A four-byte CRC (Cyclic Redundancy Code) calculated
on the preceding bytes in the chunk, including the chunk type field and
chunk data fields, but not including the length field." is confusing.
What is says is that you really only take the type field and the data,
as far as I understand it anyway, so saying "including" the only things
you include is misleading.

svgeesus:
Normative ref to RFC1952 for CRC algorithm. Sample code is informative, appendix is not.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2198,9 +2198,9 @@ algorithm</h2>
 <p>CRC fields are calculated using standardized CRC methods with
 pre and post conditioning, as defined by [[ISO 3309]]
 and [[ITU-T V.42]].
-The CRC polynomial employed,
+The CRC polynomial employed—
 which is identical to that used in the
-GZIP file format specification [[RFC1952]]
+GZIP file format specification [[RFC1952]]—
 is</p>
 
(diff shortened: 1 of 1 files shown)


ProgramMax (approved):
LGTM.

Although, we should probably use the em dash (—) instead of comma splicing.
The em dash—which I think is lovely—lets you interject thoughts. (Like that example.)

PR #161, opened 2022-08-10
Fix informative ref to Kasson

svgeesus:
The reference incorrectly was `[[??Kasson]]`, should be `[[?Kasson]]`


ProgramMax:
Good catch. Post-facto LGTM.

PR #164, opened 2022-08-11
biblio must use authors, which is an array; fix #160

### LINKED ISSUE #160: References section is gone!
Not sure when or how it happened, but while preparing to make  a spec edit I noticed we no longer have the [autogenerated References section](https://respec.org/docs/#references-0)

image

There are three ReSpec errors, I don't know if they are related though.

svgeesus:


ProgramMax (approved):
LGTM

PR #165, opened 2022-08-20
Add missing chunk ordering diagrams (APNG), fix #143

### LINKED ISSUE #143: Missing chunk ordering diagrams (APNG)
The chunk ordering lattice diagrams in [Chunk Ordering](http://localhost:8000/index.html#5ChunkOrdering) are for statinc PNG and so do not include the APNG chunks.

Copied and edited versions are needed for:

- APNG with PLTE, first frame is IDAT
- APNG with PLTE, first frame is not IDAT
- APNG with no PLTE, first frame is IDAT
- APNG with no PLTE, first frame is not IDAT

svgeesus:
Adds four diagrams:

 -  APNG with PLTE, first frame is IDAT
 -  APNG with PLTE, first frame is not IDAT
 -  APNG with no PLTE, first frame is IDAT
 -   APNG with no PLTE, first frame is not IDAT

Fixes one other diagram which still showed a bounding box rectangle used during editing

Adds these diagrams to the spec.

The figure elements have sensible ids. The object elements have ids, which are at least unique, but I would also be fine with dropping the object ids.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2504,14 +2504,6 @@ type="image/svg+xml">
 class="chunk">PLTE</span></a></figcaption>
 </figure>
 
-<!--
-for animated images where the static image forms the first frame in
-<a href="#lattice-apng-static-with-plte"></a> and
-<a href="#lattice-apng-static-without-plte"></a>, and
-for animated images where the static image is not part of the animation in
-<a href="#lattice-apng-nostatic-with-plte"></a> and
-<a href="#lattice-apng-nostatic-without-plte"></a>. -->
-
 
 <figure id="lattice-apng-static-with-plte">
 <object id="figure52a" data="figures/lattice-diagram-apng-static-first-with-plte.svg" type="image/svg+xml">
--- a/index.html
+++ b/index.html
@@ -2229,9 +2229,16 @@ table to accelerate the computation. See <a href=
 
 <p>The constraints on the positioning of the individual chunks
 are listed in <a href="#table53">
-</a> and illustrated diagrammatically in <a href=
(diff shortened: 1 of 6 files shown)


index.html
```diff
@@ -2494,9 +2501,50 @@ <h2>Chunk ordering</h2>
...
 </figure>
 
+<!--
+for animated images where the static image forms the first frame in
+<a href="#lattice-apng-static-with-plte"></a> and
+<a href="#lattice-apng-static-without-plte"></a>, and
+for animated images where the static image is not part of the animation in
+<a href="#lattice-apng-nostatic-with-plte"></a> and
+<a href="#lattice-apng-nostatic-without-plte"></a>. -->
+
```
ProgramMax, line 2514: I think this was copied & pasted in error.

ProgramMax (approved):
Mostly LGTM.

I agree that the objects probably shouldn't have IDs.

I think there might have been a temporary copy & paste that wasn't intended to be included. Let me know if that was intentional.

Additionally, I'm a little tempted to make all those diagrams the same but remove only the key pieces from each. That way a gap from the missing piece helps illustrate what is different between the diagrams. But that isn't really necessary so if we want to do that, we can do it in a separate pull request.

I'll approve since the things I'm pointing out aren't stoppers.

svgeesus:
> Additionally, I'm a little tempted to make all those diagrams the same but remove only the key pieces from each. 

That would work, at least the "with palette" and "without palette" ones are necessarily different but within each group, yes it could be done.

Merging for now since the biggest problem (complete lack of diagrams) is fixed; will look at this again though.

PR #166, opened 2022-08-20
Recommend iTXt, fix #144

### LINKED ISSUE #144: Recommend iTXT over tEXt
Spotted in [5.7.3 Defining private chunks](https://w3c.github.io/PNG-spec/#sec-defining-private-chunks):

> A private chunk SHOULD NOT be defined merely to carry textual information of interest to a human user. Instead [tEXt](https://w3c.github.io/PNG-spec/#11tEXt) chunk SHOULD BE used and corresponding keyword SHOULD BE used and a suitable keyword defined.

Given `tEXt` is Latin-1 only (and not compressed) probably `iTXt` which is UTF-8 and optionally compressed, should be recommended instead.

(Throughout, just happened to notice in this section)

svgeesus:
Recommend using UTF-8 `iTXt` rather than Latin-1 `tEXt` for new, textual information.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2534,8 +2534,8 @@ class="chunk">PLTE</span></a> in datastream</figcaption>
     experimental use.</p>
 
     <p>A private chunk SHOULD NOT be defined merely to carry textual information
-    of interest to a human user. Instead <a href="#11tEXt"><span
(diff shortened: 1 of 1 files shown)


ProgramMax (approved):
LGTM

PR #167, opened 2022-08-20
Don't recommend redundant gAMA, fix #151

### LINKED ISSUE #151: Still recommend redundant `gAMA`?
From [4.3 Colour spaces](https://w3c.github.io/PNG-spec/#4Concepts.ColourSpaces):

> It is recommended that explicit gamma information also be provided when either the first, second or third method is used, for use by PNG decoders that do not support CICP metadata, full ICC profiles or the sRGB colour space. 

Two reasons to not recommend  redundant `gAMA` when a better method is used:

1. Only a simple power function, not piecewise, so provides only an approximation of sRGB, PQ, HLG, Adobe RGB and ProPhoto transfer functions, for example
2. Spec says to ignore it, either sometimes or always. See:
  2.1. [cICP](https://w3c.github.io/PNG-spec/#11cICP) 

  > When the cICP chunk is present, decoders that recognize it SHALL ignore the following chunks:
    iCCP
    gAMA
    cHRM
    sRGB

  2.2 [iCCP]()

  > When the iCCP chunk is present, PNG decoders that recognize it and are capable of colour management shall ignore the [gAMA](https://w3c.github.io/PNG-spec/#gama-image-gamma) and [cHRM](https://w3c.github.io/PNG-spec/#chrm-primary-chromaticities-and-white-point) chunks and use the iCCP chunk instead and interpret it according to [[ICC](https://w3c.github.io/PNG-spec/#bib-icc)]. PNG decoders that are used in an environment that is incapable of full-fledged colour management should use the [gAMA](https://w3c.github.io/PNG-spec/#gama-image-gamma) and [cHRM](https://w3c.github.io/PNG-spec/#chrm-primary-chromaticitieis-and-white-point) chunks if present.

  2.3 [sRGB]()

  > When the sRGB chunk is present, it is recommended that decoders that recognize it and are capable of colour management ignore the [gAMA](https://w3c.github.io/PNG-spec/#gama-image-gamma) and [cHRM](https://w3c.github.io/PNG-spec/#chrm-primary-chromaticities-and-white-point) chunks and use the sRGB chunk instead. Decoders that recognize the sRGB chunk but are not capable of colour management are recommended to ignore the [gAMA](https://w3c.github.io/PNG-spec/#gama-image-gamma) and [cHRM](https://w3c.github.io/PNG-spec/#chrm-primary-chromaticities-and-white-point) chunks, and use the values given above as if they had appeared in [gAMA](https://w3c.github.io/PNG-spec/#gama-image-gamma) and [cHRM](https://w3c.github.io/PNG-spec/#chrm-primary-chromaticities-and-white-point) chunks.

  In other words, the [value for `gAMA` when used with `sRGB` is fixed](https://w3c.github.io/PNG-spec/#table-gama-and-chrm-values-for-srgb), could be inferred, and so the explicit `gAMA` adds no value but could be incorrect.

I feel that `gAMA` (and `cHRM`) are legacy chunks; they should still be supported but we should revisit recommending adding them when better info is available in other chunks.

svgeesus:
No longer recommend adding redundant `gAMA` and `cHRM` when better methods like CICP, ICC, or sRGB are used: the description for these chunks says to ignore it anyway. These are now legacy chunks.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1214,21 +1214,18 @@ image and the reference white point.</li>
 </ol>
 
 <p>For high-end applications the first two methods provides the most
-flexibility and control. The third method enables one particular
-colour space to be indicated. The fourth method enables the exact
+flexibility and control. The third method enables one particular,
+but extremely common,
+colour space to be indicated. The fourth method,
+which was standardized before ICC profiles were widely adopted,
+enables the exact
 chromaticities of the RGB data to be specified, along with the
 gamma correction (the power function relating the desired display
 output with the image samples) to be applied (see <a
-href="#C-GammaAppendix"></a>). It is recommended that explicit gamma
-information also be provided when either the first, second or third
-method is used, for use by PNG decoders that do not support
-CICP metadata, full
-ICC profiles or the sRGB colour space. Such PNG decoders can
-still make sensible use of gamma information. PNG decoders are
-strongly encouraged to use this information, plus information
-about the display system, in order to present the image to the
-viewer in a way that reproduces as closely as possible what the image's original author
-saw.</p>
+href="#C-GammaAppendix"></a>).
+However, colour-aware applications will prefer one of the first three methods,
+while colour-unaware applications will typically ignore all four methods.
+</p>
 
 <p>Gamma correction is not applied to the alpha channel, if
 present. Alpha samples always represent a linear fraction of full


index.html
```diff
@@ -1214,21 +1214,18 @@ <h2>Colour spaces</h2>
 </ol>
 
 <p>For high-end applications the first two methods provides the most
-flexibility and control. The third method enables one particular
-colour space to be indicated. The fourth method enables the exact
+flexibility and control. The third method enables one particular,
+but extremely common,
+colour space to be indicated. The fourth method,
+which was standardized before ICC profiles were widely adopted,
+enables the exact
```
ProgramMax, line 1221: We should probably use the em dash (—) here, too.

ProgramMax (approved):
LGTM

Although, we should use the em dash. It isn't critical so I'll approve. But we should fix it sooner or later.

PR #172, opened 2022-08-30
Oor palette index, fix #96

### LINKED ISSUE #96: Out-of-range palette indices
For indexed color images the following rule applies to PLTE:

[11.2.3 PLTE Palette](https://www.w3.org/TR/2003/REC-PNG-20031110/#11PLTE)

> ... It is permissible to have fewer entries than the bit depth would allow. In that case, any out-of-range pixel value found in the image data is an error.

By default libpng does not handle out-of-range values as an error, instead they're interpreted as black pixels. If the output format has an alpha channel the maximum alpha value (fully opaque) is assumed.

The first attached image is a 128x128 8-bit indexed color image with gray borders, the 3 palette entries are: white, gray and blue in that order.
Only the gray and blue entries are used from the palette but black shows up in the image which is not part of the palette. The bottom half is blue and the top half is filled with invalid values (`100`), this part is black even though the first palette entry is white, it is not defaulting to the first entry or anything like that.

index

The second image is the same but a tRNS chunk was added with 3 entries, the alpha values are 0, 255, 0 meaning the blue part should be transparent (and it is). This is just to show that when an alpha channel is added the part with out-of-range values will always be black and opaque, regardless of transparency values.

index

Should this be standardized to prevent more alternative and incompatible behaviors? Instead of saying it's an error it could be a choice between treating them as black, opaque pixels OR as an error.

svgeesus:
Defines opaque black as one common error recovery method.
Decoders have varied, so encoders must not rely on this presentation.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6517,6 +6517,13 @@ need not terminate.</p>
 warning to the user if appropriate, recover from the error, and
 continue processing normally.</p>
 
+<p>When decoding an indexed-color PNG, if out-of-range indexes are encountered,
+  decoders have historically varied in their handling of this error.
+  Displaying the pixel as opaque black is one common error recovery tactic,
+  although such behaviour is not required by this specification
+  and must not be relied on by encoders.
+</p>
+
 <p>Decoders that do not compute CRCs should interpret apparent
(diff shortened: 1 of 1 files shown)


ProgramMax (approved):
LGTM

PR #175, opened 2022-09-20
Move md-style internal links to HTML markup

palemieux:


index.html
```diff
@@ -1568,38 +1568,38 @@ <h2>Chunk types</h2>
   which encoders may generate and decoders may interpret.</p>
 
   <!-- <ol start="5"> --><ol>
-  <li>Transparency information: <a href="#trns-transparency"><span class=
-  "chunk">tRNS</span></a> (see [[#transparency-information]]).</li>
+  <li>Transparency information: <a href="#11tRNS"><span class=
```
ProgramMax, line 1571: I'm mildly curious where #trns-transparency came from. It isn't currently in the doc.
But not curious enough to investigate. :)
The fix is correct no matter how we got here.

ProgramMax (approved):
LGTM

index.html
ProgramMax: I realized all these broken fragments are probably when I tried to remove hard-coded section numbers in fragments. We decided to instead keep the fragment names to maintain historical inbound links.

Sorry about that.

Also, I thought we had already fixed them. I'm surprised I had left so many broken.
Anyway, thanks for finding and fixing these. :)

PR #178, opened 2022-09-23
Clean-up colour type definition

palemieux:
I had to use best judgement in some cases.

There seem to have been a conscious decision to differentiate between the image type (`Greyscale`) and the corresponding value of the `colour type` field (0). Sometimes both are present together, e.g. Table 8. I suggest we keep this unchanged.

### DIFF
--- a/index.html
+++ b/index.html
@@ -458,16 +458,6 @@ Each chunk is either a
 <a>ancillary chunk</a>.
 </dd>
 
-<!-- Maintain a fragment named "3colourType" to preserve incoming links to it -->
-<dt id="3colourType"><dfn>colour type</dfn></dt>
-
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -5388,7 +5377,7 @@ <h2><span class="chunk">fdAT</span>
       including the filter byte at the beginning of each scanline,
       similar to the uncompressed data
       of all the <span class="chunk">IDAT</span> chunks.
-      It utilizes the same bit depth, color type,
+      It utilizes the same bit depth, <a>colour type</a>,
```
ProgramMax, line 5380: Thank you for catching this. :)

ProgramMax (approved):
LGTM but I would like to preserve the comment.

PR #179, opened 2022-09-25
Clean-up filter definition

palemieux:


palemieux:
@ProgramMax Is there a difference between _The encoder shall use only a single filter method for an interlaced PNG image_ in 7 and _All the reduced images in an interlaced image shall use a single filter method_ in 9.1?

ProgramMax:
I see this but probably will not be able to get to it until tomorrow.

A quick response on interlaced & reduced images in an interlaced image: I have to double check but I *believe* a "reduced image" is basically the intermediary step images of the interlacing. Unlike TV interlacing, PNG interlacing works in cubes and has multiple levels (not just odd field, even field). As a result, the filter (which typically applies to an entire scanline) doesn't quite make sense in this context. So rather than allow a new filter for a new scanline, it is just one filter for all reduced images.

I *think*. I could be totally wrong about all that. :) I'll have to read up.

palemieux:
Ok. Hopefully _The encoder shall use only a single filter method for an interlaced PNG image_ also implies _All the reduced images in an interlaced image shall use a single filter method_, so we can get rid of the latter.

ProgramMax (approved):
LGTM

PR #201, opened 2022-10-04
Clean-up frame buffer definition

palemieux:


index.html
```diff
@@ -6920,7 +6920,7 @@ <h2>Decoder gamma
 calibration measurements are available for the particular <a>CRT</a>
 used.</p>
 
-<p>Many modern frame buffers have lookup tables that are used to
+<p>Many modern <a>frame buffer</a>s have lookup tables that are used to
```
ProgramMax, line 6923: I think ReSpec is okay with plurals. If you haven't tried, can you see if including the 's' is fine?

ProgramMax (changes requested):
LGTM, but try a change for me

index.html
palemieux: Thanks for the catch!

palemieux:
```suggestion
<p>Many modern <a>frame buffers</a> have lookup tables that are used to
```

ProgramMax (approved):
LGTM

PR #209, opened 2022-10-05
Clean-up gamma definition

palemieux:
Added definition of transfer function and gamma value

### DIFF
--- a/index.html
+++ b/index.html
@@ -1395,7 +1395,7 @@ if no better option is available.</td>
 
 <tr>
 <td>Gamma and chromaticity</td>
-<td><a>Gamm value</a> of the image with respect to the desired
+<td><a>Gamma value</a> of the image with respect to the desired
 output intensity, and <a>chromaticity</a> characteristics of the RGB
 values used in the image.</td>
 </tr>
--- a/index.html
+++ b/index.html
@@ -764,7 +764,7 @@ image it is the same as the <a>bit depth</a>.</dd>
   </dd>
 
 <dt><dfn>transfer function</dfn></dt>
-<dd>function relating the desired display output with the image samples</dd>
+<dd>function relating image luminance with image samples</dd>
 
 <!-- Maintain a fragment named "3truecolour" to preserve incoming links to it -->
 <dt id="3truecolour"><dfn>truecolour</dfn></dt>
--- a/index.html
+++ b/index.html
@@ -323,7 +323,7 @@ supported, with optional transparency. Sample depths range from 1
 to 16 bits. PNG is fully streamable with a progressive display
 option. It is robust, providing both full file integrity checking
 and simple detection of common transmission errors. PNG can store
-gamma and chromaticity data as well as a full ICC colour profile
+<a>gamma value</a> and chromaticity data as well as a full ICC colour profile
 or CICP metadata,
 for accurate colour matching on heterogenous platforms. This
 Standard defines the Internet Media types "image/png"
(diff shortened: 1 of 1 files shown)


palemieux:
@ProgramMax @svgeesus Please review carefully _gamma_ was used pretty loosely

index.html
```diff
@@ -1401,7 +1395,7 @@ <h2>Additional
 
 <tr>
 <td>Gamma and chromaticity</td>
-<td>Gamma characteristic of the image with respect to the desired
+<td><a>Gamm value</a> of the image with respect to the desired
```
svgeesus, line 1398:
```suggestion
<td><a>Gamma value</a> of the image with respect to the desired
```

index.html
```diff
@@ -516,20 +516,12 @@ <h2 id="3Definitions">Definitions</h2>
...
-<dd>exponent that describes approximations to certain non-linear
-transfer functions encountered in image capture and reproduction.
-Within this specification, gamma is the exponent in the
-transfer function from <tt>display_output</tt> to
-<tt>image_sample</tt>
-<pre>
-<tt>image_sample = display_output<sup>gamma</sup></tt>
-</pre>
-where both <tt>display_output</tt> and <tt>image_sample</tt>
-are scaled to the range 0 to 1.
```
svgeesus, line 531: Seems important to still point out that component values are in the range [0..1] when the power law is applied.(Not [0..255] for example).

index.html
```diff
@@ -767,6 +759,9 @@ <h2 id="3Definitions">Definitions</h2>
     will display this rather than the animation.
   </dd>
 
+<dt><dfn>transfer function</dfn></dt>
+<dd>function relating the desired display output with the image samples</dd>
```
svgeesus, line 763: This is true for display-referrred color spaces, but not for scene-referred color spaces.

index.html
palemieux: This is detailed elsewhere in the document, e.g. at 13.12.

Also, a definition should be written such that it can replace the term in its context.

index.html
```diff
@@ -516,20 +516,12 @@ <h2 id="3Definitions">Definitions</h2>
 
 <dd>the red, green, blue and alpha components are all set to zero.</dd>
 
+<dt><dfn>gamma value</dfn></dt>
+<dd>value of the exponent of a <a>gamma</a> <a>transfer function</a></dd>
```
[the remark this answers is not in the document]
palemieux: > do we say that a gamma transfer function is a type of power-law transfer function?

Yes.

> I see both terms here, but don't see one referencing the other.

gamma value references gamma, which references transfer function.

index.html
palemieux: What about _function relating image luminance with image samples_

index.html
```diff
@@ -3634,10 +3628,7 @@ <h2><span class="chunk">gAMA</span>
 103 65 77 65
 </pre>
 
-<p>The <span class="chunk">gAMA</span> chunk specifies the
-relationship between the image samples and the desired display
-output intensity. Gamma is defined in <a href=
-"#3gamma">gamma</a>.</p>
+<p>The <span class="chunk">gAMA</span> chunk specifies a <a>gamma value</a>.</p>
```
ProgramMax, line 3631: -span
+`<a class="chunk" href="#11gAMA">`

ProgramMax (changes requested):

index.html
palemieux: Fixed at https://github.com/w3c/PNG-spec/pull/209/commits/52138fa0f137adaadd7c80e1dc28e19c16741b90

index.html
palemieux: See https://github.com/w3c/PNG-spec/pull/209/commits/b3c296d4b98baa4b0f2c74ee18cd5c71338f612e

ProgramMax:
My concern has been addressed. LGTM once @svgeesus's concerns are also quelled.

palemieux:
@svgeesus Ping?

svgeesus (approved):
Looks good now

svgeesus:
@ProgramMax looks like it needs an explicit approve from you

ProgramMax (approved):
LGTM

PR #222, opened 2022-10-08
Remove the definition for PNG four-byte signed integer

### LINKED ISSUE #217: Are PNG four-byte signed integer used anywhere in the specification?
I could not find any evidence.

palemieux:
Closes #217

### DIFF
--- a/index.html
+++ b/index.html
@@ -673,16 +673,6 @@ rules, even for unknown chunk types.</dd>
 
 <dd><a>PNG datastream</a> stored as a file.</dd>
 
-<!-- Maintain a fragment named "3PNGfourByteSignedInteger" to preserve incoming links to it -->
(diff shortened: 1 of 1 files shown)


ProgramMax (approved):
LGTM.

I should call out, this loses a fragment which we were otherwise preserving. But I think this is the right move.

Maintaining fragments to keep inbound links makes sense when the content is still there. But since this content is removed, the fragment cannot be placed correctly, anyway.

palemieux:
> I should call out, this loses a fragment which we were otherwise preserving. But I think this is the right move.

@ProgramMax We could always add a paragraph somewhere that says "the contents at this link have been removed"?

ProgramMax:
We could do that. I don't think we need it, though.

If we did add it, maybe somewhere in the appendix. But then are we going to maintain a list of everything we deleted?
We do maintain a list of changes since last edition. But that has limited size. This would just keep growing. So I'm okay without it.

PR #226, opened 2022-10-12
Additional fixes to description of Narrow Range

digitaltvguy:
Another incorrect description of reference white was corrected.


ProgramMax (approved):
Although this seems like a simple change, it is about a subject I know nothing about.
I suppose I approve of the change. You know about this and I trust you.

But ideally another person who does know about this would approve the change.

svgeesus (approved):
This matches my understanding of why narrow range is used

PR #235, opened 2022-11-01
fix ascii-art figures, fix #84

### LINKED ISSUE #84: ASCII-art diagrams
ASCII-art diagrams are not very accessible, and should really be replaced with SVG diagrams and good alt text, or in some cases tables would be better

An example:

```html
<pre>
   cHNk  &lt;-- 32 bit chunk type represented in text form
   ||||
   |||+- Safe-to-copy bit is 1 (lower case letter; bit 5 is 1)
   ||+-- Reserved bit is 0     (upper case letter; bit 5 is 0)
   |+--- Private bit is 0      (upper case letter; bit 5 is 0)
   +---- Ancillary bit is 1    (lower case letter; bit 5 is 1)
</pre>
```

svgeesus:
image

image

### DIFF
--- a/figures/chunk-bits.svg
+++ b/figures/chunk-bits.svg
@@ -1,0 +1,61 @@
+<?xml version="1.0"?>
+
+<svg viewBox="0 0 436 200" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+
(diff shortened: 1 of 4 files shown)


ProgramMax:
That depth scaling example image is really well done.

ProgramMax (approved):
LGTM

PR #243, opened 2022-11-22
Update README.md

svgeesus:
Bring up to date


ProgramMax:
LGTM.
I'll update the format checker to only complain about new error instead of mark your pull request as bad because of old, existing behaviors.

ProgramMax (approved):

PR #247, opened 2022-12-02
Clarify the description of the language tag field of the iTXt chunk

### LINKED ISSUE #241: iTXt language tag description needs attention
11.3.3.4 iTXt International textual data
https://www.w3.org/TR/2022/WD-png-3-20221025/#11iTXt

> The language tag defined in [[BCP47](https://www.w3.org/TR/2022/WD-png-3-20221025/#bib-bcp47)] indicates the human language used by the translated keyword and the text. Unlike the keyword, the language tag is case-insensitive. It is an ISO 646.IRV:1991 [[ISO646](https://www.w3.org/TR/2022/WD-png-3-20221025/#bib-iso646)] string consisting of hyphen-separated words of 1-8 alphanumeric characters each (for example cn, en-uk, no-bok, x-klingon, x-KlInGoN). If the first word is two or three letters long, it is an ISO language code [[BCP47](https://www.w3.org/TR/2022/WD-png-3-20221025/#bib-bcp47)]. If the language tag is empty, the language is unspecified.

There are several issues with the quoted text.

None of the language tag examples are correct! There is no primary language subtag `cn` (this should probably be `zh`?). The correct country code for the UK is `GB`, so the tag should be `en-gb`. There is no extended language subtag `bok`: Norwegian Bokmal should be tagged `nb`. Both Klingon tags are "valid" from the point of view that they are private use. However, Klingon has an ISO 693-3 code that is in the IANA language subtag registry (`tlh`).

The phrase *If the first word* is odd. It should probably use BCP47's term (subtag). Referencing ISO language codes doesn't make sense vs. mentioning the [LSTR](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry).

I would suggest:

> The language tag is a "well-formed" language tag defined by [\[BCP47\]](https://rfc-editor.org/bcp/bcp47). It consists of an ASCII string of hyphen-separated "subtags", each of which is 1-8 alphanumeric characters long. Unlike the keyword, the language tag is case-insensitive. Subtags must appear in the IANA language subtag registry. If the language tag is empty, the language is unspecified. Examples of language tags include: `en`, `en-GB`, `es-419`, `zh-Hans`, `zh-Hans-CN`, `tlh-Cyrl-AQ`, `ar-AE-u-nu-latn`, `x-private`. 

For more info, see our Best Practices [here](https://www.w3.org/TR/international-specs/#lang_values)

palemieux:
Closes #241

### DIFF
--- a/index.html
+++ b/index.html
@@ -4094,11 +4094,11 @@ with these exceptions:
           is 0 (<a>zlib</a> datastream with <a>deflate</a> compression, see <a href='#10CompressionOtherUses'></a>). For
           uncompressed text, encoders shall set the compression method to 0, and decoders shall ignore it.</p>
 
-          <p>The language tag defined in [[BCP47]] indicates the human language used by the translated keyword and the text. Unlike
-          the keyword, the language tag is case-insensitive. It is an ISO 646.IRV:1991 [[ISO646]] string consisting of
-          hyphen-separated words of 1-8 alphanumeric characters each (for example cn, en-uk, no-bok, x-klingon, x-KlInGoN). If the
-          first word is two or three letters long, it is an ISO language code [[BCP47]]. If the language tag is empty, the language
-          is unspecified.</p>
+          <p>The language tag is a well-formed language tag defined by [[BCP47]]. Unlike the keyword, the language tag is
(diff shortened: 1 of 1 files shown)


aphillips (approved):
Looks good to me.

PR #251, opened 2022-12-19
Use hex for character codes, Unicode character names. Fix #242

### LINKED ISSUE #242: references to characters and character values
Keywords and text string
https://www.w3.org/TR/2022/WD-png-3-20221025/#11keywords

> Keywords shall contain only printable Latin-1 [[ISO_8859-1](https://www.w3.org/TR/2022/WD-png-3-20221025/#bib-iso_8859-1)] characters and spaces; that is, only character codes 32-126 and 161-255 decimal are allowed. To reduce the chances for human misreading of a keyword, leading spaces, trailing spaces, and consecutive spaces are not permitted in keywords, nor is the non-breaking space (code 160) since it is visually indistinguishable from an ordinary space.

Hex is preferred to decimal values when referring to code point ranges and I18N prefers if references also use the Unicode name when referring to specific characters. The Latin-1 heritage of some parts of PNG should not prevent switching to this best practice. 

For example, instead of *the non-break space (code 160)* it should say *U+00A0 NON-BREAKING SPACE* and the ranges "32-126" and "161-255" would be `0x20-7E` and `0xA1-FF` respectively. This is easier for folks familiar with Unicode ;-).

svgeesus:
This does not change the PNG chunk signatures from decimal to hex, which is unrelated to character encoding so should be a new issue

### DIFF
--- a/index.html
+++ b/index.html
@@ -3417,7 +3417,7 @@ with these exceptions:
           </table>
 
           <p>The profile name may be any convenient name for referring to the profile. It is case-sensitive. Profile names shall
-          contain only printable Latin-1 characters and spaces (only character codes 32-126 and 161-255 decimal are allowed).
+          contain only printable Latin-1 characters and spaces (only character codes 0x20-7E and 0xA1-FF are allowed).
           Leading, trailing, and consecutive spaces are not permitted. The only compression method defined in this specification is
           method 0 (<a>zlib</a> datastream with <a>deflate</a> compression, see <a href='#10CompressionOtherUses'></a>). The
           compression method entry is followed by a compressed profile that makes up the remainder of the chunk. Decompression of
@@ -3924,9 +3924,9 @@ with these exceptions:
 
           <p>Keywords of general interest SHOULD be listed in [[PNG-EXTENSIONS]].</p>
 
-          <p>Keywords shall contain only printable Latin-1 [[ISO 8859-1]] characters and spaces; that is, only character codes
-          32-126 and 161-255 decimal are allowed. To reduce the chances for human misreading of a keyword, leading spaces, trailing
-          spaces, and consecutive spaces are not permitted in keywords, nor is the non-breaking space (code 160) since it is
+          <p>Keywords shall contain only printable Latin-1 [[ISO 8859-1]] characters and spaces; that is, only character codes 0x20-7E and 0xA1-FF are allowed. To reduce the chances for human misreading of a keyword, leading spaces, trailing
+          spaces, and consecutive spaces are not permitted in keywords, nor is 
+          U+00A0 NON-BREAKING SPACE since it is
           visually indistinguishable from an ordinary space.</p>
 
           <p>Keywords shall be spelled exactly as registered, so that decoders can use simple literal comparisons when looking for
@@ -4381,7 +4381,7 @@ with these exceptions:
 
           <p>The palette name is case-sensitive, and subject to the same restrictions as the keyword parameter for the <a class=
           "chunk" href="#11tEXt">tEXt</a> chunk. Palette names shall contain only printable Latin-1 characters and spaces (only
-          character codes 32-126 and 161-255 decimal are allowed). Leading, trailing, and consecutive spaces are not permitted.</p>
+          character codes 0x20-7E and 0xA1-FF are allowed). Leading, trailing, and consecutive spaces are not permitted.</p>
 
           <p>The <span class="chunk">sPLT</span> sample depth shall be 8 or 16.</p>
 
@@ -5632,7 +5632,7 @@ scale_factor_Y = max(1.0, display_ratio/image_ratio)</code>
 
       <p>Decoders running on systems with non-Latin-1 character set encoding should provide character code remapping so that
       Latin-1 characters are displayed correctly. Some systems may not provide all the characters defined in Latin-1. Mapping
-      unavailable characters to a visible notation such as "<code>\nnn</code>" is a good fallback. Character codes 127-255 should
+      unavailable characters to a visible notation such as "<code>\nnn</code>" is a good fallback. Character codes 0x7F-FF should
       be displayed only if they are printable characters on the decoding system. Some systems may interpret such codes as control
       characters; for security, decoders running on such systems should not display such characters literally.</p>
 


index.html
```diff
@@ -5632,7 +5632,7 @@ <h2>Text chunk processing</h2>
 
       <p>Decoders running on systems with non-Latin-1 character set encoding should provide character code remapping so that
       Latin-1 characters are displayed correctly. Some systems may not provide all the characters defined in Latin-1. Mapping
-      unavailable characters to a visible notation such as "<code>\nnn</code>" is a good fallback. Character codes 127-255 should
+      unavailable characters to a visible notation such as "<code>\nnn</code>" is a good fallback. Character codes 0x7F-FF should
       be displayed only if they are printable characters on the decoding system. Some systems may interpret such codes as control
```
aphillips, line 5636: This text is slightly mutant, since it confuses "character codes" with byte values. I would probably rewrite this paragraph thusly:

> Decoders running on systems with a non-Latin-1 [legacy character encoding](https://www.w3.org/TR/i18n-glossary/#def_legacy_character_encoding) should remap character codes so that Latin-1 characters are displayed correctly. Unsupported characters should be replaced with a system-appropriate replacement character (such as U+FFFD REPLACEMENT CHARACTER or U+001A SUB) or mapped to a visible notation such as `\nnn`. Characters should only be displayed if they are printable characters on the decoding system. Some byte values may be interpreted by the decoding system as control characters.

(Note that U+003F QUESTION MARK is a common replacement character in many encodings)

svgeesus:
@aphillips does this look good?

index.html
```diff
@@ -3417,7 +3417,7 @@ <h2><span class="chunk">iCCP</span> Embedded ICC profile</h2>
           </table>
 
           <p>The profile name may be any convenient name for referring to the profile. It is case-sensitive. Profile names shall
-          contain only printable Latin-1 characters and spaces (only character codes 32-126 and 161-255 decimal are allowed).
+          contain only printable Latin-1 characters and spaces (only character codes 0x20-7E and 0xA1-FF are allowed).
```
aphillips, line 3420: Do we want to say "character codes" or "byte values"? Unlike most legacy encodings, Latin-1 and ASCII map directly to the first 256 code points of Unicode, so it is convenient to talk in terms of Unicode's jargon here vs. having to talk about bytes => characters => code points. That is, you could say:

> ...spaces (only code points U+0020 - U+007E and U+00A1 - U+00FF are allowed)

Failing that (and being consistent with the current text) I'd probably write:

> ...spaces (that is, only byte values in the ranges 0x20-7E and 0xA1-FF are allowed).

index.html
```diff
@@ -3924,9 +3924,9 @@ <h2>Keywords and text strings</h2>
 
           <p>Keywords of general interest SHOULD be listed in [[PNG-EXTENSIONS]].</p>
 
-          <p>Keywords shall contain only printable Latin-1 [[ISO 8859-1]] characters and spaces; that is, only character codes
-          32-126 and 161-255 decimal are allowed. To reduce the chances for human misreading of a keyword, leading spaces, trailing
-          spaces, and consecutive spaces are not permitted in keywords, nor is the non-breaking space (code 160) since it is
+          <p>Keywords shall contain only printable Latin-1 [[ISO 8859-1]] characters and spaces; that is, only character codes 0x20-7E and 0xA1-FF are allowed. To reduce the chances for human misreading of a keyword, leading spaces, trailing
```
aphillips, line 3927: Same comment as above.

aphillips (commented):
Looks pretty good. Some minor comments.

index.html
svgeesus: Fixed as suggested (including U+003F QUESTION MARK)

index.html
svgeesus: Changed to "code points"

index.html
svgeesus: and here

index.html
```diff
@@ -4381,7 +4381,7 @@ <h2><span class="chunk">sPLT</span> Suggested palette</h2>
 
           <p>The palette name is case-sensitive, and subject to the same restrictions as the keyword parameter for the <a class=
           "chunk" href="#11tEXt">tEXt</a> chunk. Palette names shall contain only printable Latin-1 characters and spaces (only
-          character codes 32-126 and 161-255 decimal are allowed). Leading, trailing, and consecutive spaces are not permitted.</p>
+          character codes 0x20-7E and 0xA1-FF are allowed). Leading, trailing, and consecutive spaces are not permitted.</p>
```
aphillips, line 4384: character codes => code points?

index.html
```diff
@@ -5630,11 +5630,17 @@ <h2>Text chunk processing</h2>
       rely on this; it is best to recognize all the common newline combinations (CR, LF, and CR-LF) and display each as a single
       newline. TAB can be expanded to the proper number of spaces needed to arrive at a column multiple of 8.</p>
 
-      <p>Decoders running on systems with non-Latin-1 character set encoding should provide character code remapping so that
-      Latin-1 characters are displayed correctly. Some systems may not provide all the characters defined in Latin-1. Mapping
-      unavailable characters to a visible notation such as "<code>\nnn</code>" is a good fallback. Character codes 127-255 should
-      be displayed only if they are printable characters on the decoding system. Some systems may interpret such codes as control
-      characters; for security, decoders running on such systems should not display such characters literally.</p>
+      <p>Decoders running on systems with non-Latin-1 <a>legacy character encoding</a> should remap character codes so that
```
aphillips, line 5633: systems with non-Latin-1 => systems with *a* non-Latin-1 ?

index.html
```diff
@@ -3417,7 +3417,7 @@ <h2><span class="chunk">iCCP</span> Embedded ICC profile</h2>
           </table>
 
           <p>The profile name may be any convenient name for referring to the profile. It is case-sensitive. Profile names shall
-          contain only printable Latin-1 characters and spaces (only character codes 32-126 and 161-255 decimal are allowed).
+          contain only printable Latin-1 characters and spaces (only code points 0x20-7E and 0xA1-FF are allowed).
```
aphillips, line 3420: Nit: I would probably use the Unicode `U+xxxx` notation for code points here and on line 3927, but I'm okay with this.

aphillips (commented):
Minor comments, otherwise looks good.

index.html
svgeesus: The text `0x20-7E and 0xA1-FF` is your suggestion, actually :)

index.html
svgeesus: yes

index.html
svgeesus: agreed

svgeesus:
@aphillips if this is okay, please mark your review as Approved?

aphillips (approved):
Good to go.

svgeesus:
@ProgramMax merging needs at least one WG approval too.

ProgramMax (approved):
Sorry about the delay.
I somehow missed the email where I was pinged.

LGTM. Thanks for doing this.

PR #252, opened 2022-12-19
Prefer iTXt

svgeesus:
Corrects both paragraphs per @aphillips suggestions. Adds links to I18n glossary.

ReSpec complains about these being normative.


aphillips (approved):
Looks good to me.

ProgramMax (commented):
LGTM as well.
Thank you for clearing up existing document issues, too.

PR #254, opened 2023-01-11
decimal to hex

svgeesus:
- Convert 4-byte decimal chunk signatures to hex. (Previous decimal values retained in comments for verification)
 
- Consistent introductory text. Most used
 ```html
<p>The four-byte chunk type field contains the hexadecimal values</p>
```
but a few used this style
```html
<p>The four decimal values below correspond to the four-byte gAMA chunk type field:</p>
```
Now they all use the shorter, clearer text.

 - PNG signature had both decimal and hex, now has hex only.

 - A few other occurrences of decimal values (like specifying ranges of control characters in text) also converted to hex

These are all accompanied by text stating _explicitly_ that the values are hexadecimal, so I didn't use noisy `0x4C` just `4C`

### DIFF
--- a/index.html
+++ b/index.html
@@ -283,7 +283,6 @@ with these exceptions:
 
   <h2 id="Introduction">Introduction</h2>
 
-
   <p>The design goals for this specification were:</p>
 
   <ol>
@@ -1718,14 +1717,10 @@ with these exceptions:
     <section id="5PNG-file-signature">
       <h2>PNG signature</h2>
 
-      <p>The first eight bytes of a PNG datastream always contain the following (decimal) values:</p>
+      <p>The first eight bytes of a PNG datastream always contain the following hexadecimal values:</p>
 
       <pre>
-   137 80 78 71 13 10 26 10
-</pre>
-      <p>which are (in hexadecimal):</p>
-
-      <pre>
+  <!-- 137 80 78 71 13 10 26 10 -->
   89 50 4E 47 0D 0A 1A 0A
 </pre>
       <p>This signature indicates that the remainder of the datastream contains a single PNG image, consisting of a series of
@@ -1773,8 +1768,8 @@ with these exceptions:
         <tr>
           <td>Chunk Type</td>
           <td>
-            A sequence of four bytes defining the chunk type. Each byte of a chunk type is restricted to the decimal values 65 to
-            90 and 97 to 122. These correspond to the uppercase and lowercase ISO 646 [[ISO646]] letters
+            A sequence of four bytes defining the chunk type. Each byte of a chunk type is restricted to the hexadecimal values 41
+            to 5A and 61 to 7A. These correspond to the uppercase and lowercase ISO 646 [[ISO646]] letters
             (<code>A</code>-<code>Z</code> and <code>a</code>-<code>z</code>) respectively for convenience in description and
             examination of PNG datastreams. Encoders and decoders shall treat the chunk types as fixed binary values, not character
             strings. For example, it would not be correct to represent the chunk type <a class="chunk" href="#11IDAT">IDAT</a> by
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4115,26 +4116,23 @@ <h2>Keywords and text strings</h2>
...
-            can be used to convey characters in any language. 
-            There is an option to compress text strings
-            in the <a class="chunk" href="#11iTXt">iTXt</a> chunk.
-            <a class="chunk" href="#11iTXt">iTXt</a> is recommended for all text strings, as it supports Unicode.
-            There are also <a class="chunk" href="#11tEXt">tEXt</a> and <a class="chunk" href="#11zTXt">zTXt</a> chunks,
-            whose content is restricted to the printable Latin-1 character set plus U+000A LINE FEED (LF).
-            Text strings in <a class="chunk" href="#11zTXt">zTXt</a> are compressed into <a>zlib</a> datastreams 
-            using <a>deflate</a> compression (see <a href='#10CompressionOtherUses'></a>).  </p>
-
+          <p>The <a class="chunk" href="#11iTXt">iTXt</a> chunk uses the UTF-8 encoding [[rfc3629]] and can be used to convey
```
svgeesus, line 4119: Tidy again

ProgramMax (approved):
LGTM.

I don't think we need the decimal comments. But I don't mind them being there.
Maybe they should mention they are the decimal equivalent? I could imagine someone coming along later and wondering why they're seeing different values and if it was a mistake.

Also, should we consider prefixing all hex values with 0x? It might make for a more natural read to developers. But then again, maybe it won't. I could imagine someone jumping to iCCP, seeing 69 43 43 50, and throwing those values into their code as decimal.

Last thought: HTML Tidy's lack of idempotency is frustrating to me. I feel like HTML Tidy is the least bad tool available that I've seen, but it still isn't good. Maybe we can contribute to HTML Tidy and improve it.

index.html
```diff
@@ -6885,131 +6874,140 @@ <h3 id="A-Media-type">image/png</h3>
...
-    
-          <dt>Intended usage:</dt>
-          <dd>Common</dd>
-    
-          <dt>Author/Change controller:</dt>
-          <dd>W3C</dd>
-    
-        </dl>
-      </section>
+
```
svgeesus, line 6877: Thanks, HTML Tidy

svgeesus:
> I could imagine someone jumping to iCCP, seeing 69 43 43 50, and throwing those values into their code as decimal.

I could too, if the sentence just prior didn't say "contains the hexadecimal values". But it does, every time.

svgeesus:
As someone who

 a) is supremely indifferent to the tabs vs spaces wars
 b) uses Git blame a lot
 c) uses [Semantic linefeeds](https://rhodesmill.org/brandon/2012/one-sentence-per-line/) in other groups

I find Tidy adds nothing of value (to me) and removes things of value like meaningful commit diffs. But I'm aware others have different opinions.

I find Tidy

ProgramMax:
>> I could imagine someone jumping to iCCP, seeing 69 43 43 50, and throwing those values into their code as decimal.

> I could too, if the sentence just prior didn't say "contains the hexadecimal values". But it does, every time.

Right. But the numbers stand out. If someone wasn't reading and just jumped to the numbers I could see it going wrong.

But that said,
1.) that's their own fault and they'll notice it in debugging and fix it. And
2.) if we try to cover every possible misreading case we'll never finish.

So I meant it more like a comment of "Does this cross the line enough that we should consider this misreading case?" Just making sure we've discussed it and thought it out.

My take is "69 43 43 50" has acceptable risk because of 1.) above.
"0x69 0x43 0x43 0x50" is actually harder to read, IMHO.
"0x69 43 43 50" could act as a reminder but now introduces confusion -- "Why aren't the others prefixed with 0x?"
So I think the way you wrote it is best.

PR #258, opened 2023-01-30
Added new text explaining mAst and new mLum check (MaxCLL/MaxFall)

digitaltvguy:
I hope I got this right.  It's a start.

PR #259, opened 2023-02-07
Follow-up cleanup requested by @ProgramMax

svgeesus:
A few minor things noted as needed in https://github.com/w3c/PNG-spec/pull/258


index.html
```diff
@@ -262,8 +262,9 @@
           publisher: "ITU",
           date: "2013-07",
           href: "https://www.itu.int/rec/R-REC-BT.2035-0-201307-I"
-      },
-    };
+      }
+    }
+  }
```
ProgramMax, line 267: Is this supposed to go from 2 closing braces } to 3?

Anyway, we only need to lose the comma, I think. The closing semicolon is still fine.

(Although, JavaScript tries to guess where to insert missing semicolons so maybe it will just work anyway? Probably better to be explicit about the semicolon going there.)

ProgramMax (approved):
LGTM but one potential change / suggestion

index.html
svgeesus: ReSpec would not work for me without the extra closing brace?  and VS Code gave me a syntax warning on the semicolon.

ProgramMax: *shrugs* Okay, fair enough.
I'll check again once it has landed.
LGTM

Issue #265, opened 2023-02-14
Typo in E.1 Archive Sites URL (printed)


AramZS:
Very minor issue, but spotted during my privacy review: while the linked URL is correct [the link text for the archive URL at E.1](https://www.w3.org/TR/2022/WD-png-3-20221025/#E-Archive-sites) seems to have a typo. 

It is printed as `http://www.w3.org/TR/2022/WD-PNG-20221025"/`. The quote mark preceding the trailing slash appears to be in error?

ProgramMax:
Thank you for catching that. [It looks like this has already been fixed.](https://w3c.github.io/PNG-spec/#E-Archive-sites) Should we take further action to retroactively fix that WD?

AramZS:
@ProgramMax As long as you've got it fixed for the final version, that's not an issue I'd think! Just wanted to make sure it got caught before publication.

ProgramMax:
Okay. I'll close the issue. Feel free to reopen if there is more we can do.

PR #266, opened 2023-02-14
Update links

chrisn:
Reading https://github.com/w3c/PNG-spec/issues/265, I noted that some links are still using http rather than https.

So I've been through the spec and updated links to use HTTPS URLs where available.

This change also updates some stale links, e.g., to some of the authors' personal websites, and www.zlib.org which redirects to www.zlib.net.

Some of the author links are no longer available, and I don't have an alternative to suggest, including:

- Mark Adler - http://www.alumni.caltech.edu/~madler/
- Lee Daniel Crocker - not clear if http://www.piclab.com/ belongs to him
- Robert P. Poole - http://www.users.qwest.net/~lionlad/

Should we try to update these, or leave as they are, for historical interest?


ProgramMax (approved):
LGTM

ProgramMax:
Thank you for doing this.

I don't know what the best option is for broken links.
I can understand the argument of preserving the link for historical sake.
But a broken link is misleading to users. A historical version of the spec preserves the historical link.

Anyone else have thoughts? I'm leaning toward removing broken links.

chrisn:
I agree with removing the broken links. As you say, the historical version of the spec preserves the historical links. 

So if we agree, I can update this PR. I just didn't want to go ahead and remove anything without checking.

svgeesus:
The second edition has the historical links, so broken links should be removed.

For Mark Adler though, I suggest his wikipedia page
  https://en.wikipedia.org/wiki/Mark_Adler

index.html
```diff
@@ -7468,14 +7468,14 @@ <h3>Introduction</h3>
       <h2>Archive sites</h2>
 
       <p>This specification can be found at <a href=
-      "http://www.w3.org/TR/2003/REC-PNG-20031110/index.html"><code>http://www.w3.org/TR/2003/REC-PNG-20031110/index.html</code></a>.</p>
+      "https://www.w3.org/TR/2003/REC-PNG-20031110/index.html"><code>https://www.w3.org/TR/2003/REC-PNG-20031110/index.html</code></a>.</p>
```
svgeesus, line 7471: That is old text, and now incorrect (points to second edition). It adds no value because the top of the spec also has the url, so we can just delete it.

svgeesus (approved):
LGTM

index.html
chrisn: OK, but I note that removing this means we'll have an empty section E.1

chrisn:
Thanks, I'll update the PR shortly

ProgramMax:
If section E.1 is now empty, we should probably remove it, right?
Thoughts? Follow-up PR?

chrisn:
I left it in place as the HTML contains this:

```html
<!-- Maintain a fragment named "E-Archive-sites" to preserve incoming links to it -->

<section id="E-Archive-sites">
```

But perhaps it's safe to remove anyway. I'm happy to send a follow up PR.

chrisn:
I guess there's two choices, 

* Those incoming links are important, and expect to point to "This specification can be found at...", so we should keep that text, or
* Those incoming links aren't important and so we can remove the empty section

ProgramMax:
Right.
When the user comes, they see an empty section. That's not useful to them. If we remove it and break the incoming link (which wasn't useful), they'll instead be at the top of the page.

But it could be argued that the empty section *is* useful because they can see it is empty. If they are taken to the top of the page, they might wonder "Where are the archive sites? I was expecting to see that."

Both cases make sense to me. So I'm indifferent.

chrisn:
If the Archive sites section just contains a definitive link to the spec, than sending the user to the top of the page gives them that information, which seems OK to me. I've filed https://github.com/w3c/PNG-spec/pull/276 for that change.

If on the other hand, the group prefers to keep the empty section, feel free to close the PR without merging.

Issue #263, opened 2023-02-09
Example 14 talks about minimum luminance in the max luminance section


ProgramMax:
[Example 14 ](https://w3c.github.io/PNG-spec/#example-14) talks about minimum luminance.
But this example is in the mLUm chunk which describes MaxFALL & MaxCLL.

I don't think we should mention minimums here.
Given MaxCLL and MaxFALL are content dependent, there probably aren't good default examples to use. But maybe we can list a common default max luminance for SDR and HDR just to provide some example values and how they're stored.

digitaltvguy:
The example description is wrong.  I will fix it.  The examples are real-world examples from live broadcasts.  It should be the Maximum "frame-average" light level.

digitaltvguy:
Table ID's were wrong and needed to be fixed also

ProgramMax:
Thank you, @digitaltvguy.
When you're ready, sent a pull request.
If you mention this issue the two will be linked. I usually have the commit comment say 'closes #xyz'. GitHub then knows that when the pull request is accepted, it can auto-close the issue.

digitaltvguy:
Done

svgeesus:
[Example 14](https://w3c.github.io/PNG-spec/#example-14) no longer mentions minimum luminance, so looks like this issue can be closed?

PR #267, opened 2023-02-17
Closes: Example 14 talks about minimum luminance in the max luminance section #263

digitaltvguy:
Fixed descriptions of MaxCLL and MaxFall

### DIFF
--- a/index.html
+++ b/index.html
@@ -4118,7 +4118,10 @@ with these exceptions:
            is determined by measuring each frames average light level. The frame 
            with the highest average light level becomes the MaxFALL value.  MaxCLL/MaxFall are 
            representing the entire length of content, and since PNG is most often a single frame, 
-           the start and end are the same (with the exception of image sequences).</p>
+           the start and end are the same. There are exceptions:
+                • APNG: "Animated PNG" - defined elsewhere in this document
+                • Image Sequences: A Simple set of numerically sequential frames that together represent 
+                  a single piece of content with motion video
 
   <p>The following specifies the syntax of the <span class="chunk">mLUm</span> chunk:</p>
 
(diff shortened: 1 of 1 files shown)


ProgramMax:
This pull request seems to be your previous changes + the new change.
I think you meant for the pull request to only include https://github.com/w3c/PNG-spec/pull/267/commits/5cf95d731ee4228ee6cdb16b152c85d95c9a9915

Some existing fixes are causing a conflict in this pull request.

I'll see if the conflicts can be resolved easily so we can squash this pull request and merge.

index.html
```diff
@@ -4000,7 +4000,7 @@ <h2><span class="chunk">mDCv</span> Mastering Display Color Volume</h2>
               </tr>
 
               <tr>
-                <td>0.005 cd/m<sup>2</sup></td>
+                <td>0.0005 cd/m<sup>2</sup></td>
```
ProgramMax, line 4003:
```suggestion
                <td>0.005 cd/m<sup>2</sup></td>
```

Oops.
I was able to resolve the conflicts and make this pull request clean.
However, I missed this line.
This should be 0.005, not 0.0005.

ProgramMax (changes requested):
LGTM.
Except while resolving conflicts, I accidentally left in something we don't want to change.

Can you accept the suggested change?
Then I can approve the pull request.

digitaltvguy:
Please recheck my change.  I also updated example in mDCv to show SDR display @ 203cd/m2 instead of 200 (which matches ITU BT.2408-6 pending annex).

ProgramMax (approved):
LGTM

Although, it looks like the 203 cd/m^2 change didn't get in somehow?
But we can add that in a separate pull request. This pull request is ready to go as-is.

PR #272, opened 2023-03-01
Add mlum mdvc

svgeesus:
1. Add `mLUm` to chunk ordering table
2. Update the largest with-plte and without-plte diagrams to add `mDCv` and `mLUm`
3. New versions of the other four diagrams, by deletion

_Remember the GitHub preview does not fetch webfonts so the text will not be correct_


ProgramMax (approved):
LGTM.
Goot catch on the missing mLUm.

PR #276, opened 2023-03-06
Remove empty 'Archive sites' section

chrisn:
See https://github.com/w3c/PNG-spec/pull/266 for discussion.


ProgramMax (approved):
LGTM

I am still indifferent about this change. So my "LGTM" only means the change makes sense, not that we should do it.

I defer to others to say whether they think we should do it or not.

(If enough time passes with no other comments, feel free to land this change.)

PR #280, opened 2023-03-07
Describe mLUm handling of static and animated PNG, closes #271

### LINKED ISSUE #271: Hard to understand prose for `mLUm`
The description of MaxCLL is straightforward and covers static and animated PNG:

> "Maximum Content Light Level" (MaxCLL) for the entire length of content (in every frame) 

The description of MaxFALL starts similarly

> (MaxFALL) value is determined by measuring each frame's average light level. The frame with the highest average light level becomes the MaxFALL value.

and then becomes hard to understand:

> MaxCLL/MaxFall are representing the entire length of content, and since PNG is most often a single frame, the start and end are the same.

If that is trying to say that a static PNG consists of a single frame, then that seems a convoluted way of saying it and it also seems tautological.

> There are exceptions:

I would expect the specification to say what to do in the case of those exceptions

> APNG: "Animated PNG" - defined elsewhere in this document

That doesn't seem helpful, because no-where in the APNG parts of the specification does it talk about maxFALL. Also, the correct handling has already been described: "each frame's average light level". So this wording seems to add nothing.

> Image Sequences: A Simple set of numerically sequential frames that together represent a single piece of content with motion video

This seems to be describing a set of static PNG images. That sounds like a workflow implementation detail, and in that case it is unlikely that the PNG images would contain an `mLUm` chunk.

I suggest that the clarity of the specification would be improved by simply deleting this:

> MaxCLL/MaxFall are representing the entire length of content, and since PNG is most often a single frame, the start and end are the same. There are exceptions:
> APNG: "Animated PNG" - defined elsewhere in this document
> Image Sequences: A Simple set of numerically sequential frames that together represent a single piece of content with motion video

and replacing with the more precise:

> For static PNG, the static image is analyzed and is considered to be the first (and only) frame. For animated PNG, each frame of the frame-based animation sequence is analyzed.

This uses the terminology from [Static and Animated images](https://w3c.github.io/PNG-spec/#static-and-animated-images) and also gives clarity on the case of an animated PNG where the static image is not the first frame.

svgeesus:


ProgramMax (approved):
LGTM

PR #285, opened 2023-03-22
mLUm tag changed to cLLi

digitaltvguy:
In addition, MaxCLL/MaxFALL definitions have been improved (based on Dolby description).  Also, defined expectations when there is a zero value (it means "unknown").


index.html
```diff
@@ -4099,34 +4099,36 @@ <h2><span class="chunk">mDCv</span> Mastering Display Color Volume</h2>
...
         </section>
 
-        <section id="mLUm-chunk">
-          <h2><span class="chunk">mLUm</span> Metadata for Maximum Single-Pixel and Frame-Average Luminance Levels</h2>
+        <section id="cLLi-chunk">
+          <h2><span class="chunk">cLLi</span> Content Light Level Information</h2>
           <p>The four-byte chunk type field contains the hexadecimal values</p>
 
           <pre>
 <!-- 109 76 85 109 -->6D 4C 55 6D
```
ProgramMax, line 4107: We'll need to change these values, too. But I can do that in a follow-up.

index.html
```diff
@@ -4099,34 +4099,36 @@ <h2><span class="chunk">mDCv</span> Mastering Display Color Volume</h2>
...
-          </p>
-
-           <p>The following specifies the syntax of the <span class="chunk">mLUm</span> chunk:</p>
-
-           <table id="mLUm-chunk-syntax" class="numbered simple">
-             <caption>mLUm chunk components</caption>
+          <p>If present, the <span class="chunk">cLLi</span> chunk identifies two characteristics:</p>
+             
+          <p>MaxFALL (Maximum Frame Average Light Level) indicates the maximum value of the frame 
+             average light level (in cd/m2 or nits) of the entire playback sequence. MaxFALL is
```
ProgramMax, line 4113: We should change it to `cd/m<sup>2</sup>`, but that can also happen in a follow-up.

index.html
```diff
@@ -4099,34 +4099,36 @@ <h2><span class="chunk">mDCv</span> Mastering Display Color Volume</h2>
...
-
-           <p>The following specifies the syntax of the <span class="chunk">mLUm</span> chunk:</p>
-
-           <table id="mLUm-chunk-syntax" class="numbered simple">
-             <caption>mLUm chunk components</caption>
+          <p>If present, the <span class="chunk">cLLi</span> chunk identifies two characteristics:</p>
+             
+          <p>MaxFALL (Maximum Frame Average Light Level) indicates the maximum value of the frame 
+             average light level (in cd/m2 or nits) of the entire playback sequence. MaxFALL is 
+             calculated by averaging the decoded luminance values of all the pixels within a frame.</p>
```
ProgramMax, line 4114: "within *a* frame" sounds like it is measuring a single frame.

Can we reword this? Something like "All the pixels of a given frame have their luminance values averaged. Then the maximum of those frame averages is stored." (I'm not wording this well here.)

I'm worried that "...averaging the decoded luminance values of all the pixels of all frames" sounds like one big average, instead of one max of many averages.

index.html
```diff
@@ -4099,34 +4099,36 @@ <h2><span class="chunk">mDCv</span> Mastering Display Color Volume</h2>
...
+             
+          <p>MaxCLL (Maximum Content Light Level) indicates the maximum light level of any single 
+             pixel (in cd/m2 or nits) of the entire playback sequence.  There is often an algorithmic
```
ProgramMax, line 4117: This should also be `cd/m<sup>2</sup>`. But again, that can be a follow-up.

ProgramMax (approved):
LGTM.

There might be a better way to word the MaxFALL. Let me know if you have any ideas there.

svgeesus (approved):
A few things to fix but this looks good in general

PR #301, opened 2023-04-17
Better wording on filter selection, fix #153

### LINKED ISSUE #153: Mention current state of filter selection
Section 12 mentions what filter selection has worked well in early tests: https://w3c.github.io/PNG-spec/#12Filter-selection
It also mentions "However, it is likely that better heuristics will be found as more experience is gained with PNG".

At this point, I'm sure much more experience has been gained.
We should take inventory of what different PNG encoders do for their filter selection and update this section.

svgeesus:


ProgramMax (approved):
LGTM

PR #303, opened 2023-04-19
Add links to colour image type definitions

### LINKED ISSUE #298: Unused definition of "Greyscale with alpha"
In 9a2ac909ae92872d134e81995f13389e8366f76a a new ReSpec warning arose:

'Found definition for "Greyscale with alpha", but nothing links to it. This is usually a spec bug!'

@palemieux you are perhaps best suited to handle this so I'm assigning it to you. If "Greyscale with alpha" is not referenced in the spec anywhere (Can that be right??) then perhaps we should just remove the definition? What do you think?

palemieux:
Closes #298


ProgramMax (approved):
LGTM

PR #304, opened 2023-04-27
Require Matrix Coefficients to be 0, fix #302

### LINKED ISSUE #302: Is Matrix Coefficients == 0 a requirement?
This is the text under `cICP`:

> The Matrix Coefficients value is equal to 0.
>
> NOTE
> A Matrix Coefficients value other than 0 signals a transformation between colour difference and RGB representations, which is not used in this International Standard since PNG images are RGB images.

I'm assuming this is meant to be normative, but there is no explicit shall/must. This means it could easily be missed, especially when writing compliance checkers (which typically search for all shalls/musts when adding the rules).

svgeesus:
Use suggested `shall` wording

### DIFF
--- a/index.html
+++ b/index.html
@@ -3583,7 +3583,7 @@ with these exceptions:
           <p>Each of the fields of the <span class="chunk">cICP</span> chunk corresponds to the parameter of the same name in
           [[ITU-T H.273]].</p>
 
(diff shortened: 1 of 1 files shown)


ProgramMax (approved):
LGTM

PR #305, opened 2023-04-27
Define frame, fix #286

### LINKED ISSUE #286: We need a definition of 'frame'
A definition of 'frame' should probably be in [https://w3c.github.io/PNG-spec/#apng-frame-based-animation](https://w3c.github.io/PNG-spec/#apng-frame-based-animation).

This came up in https://github.com/w3c/PNG-spec/pull/284 where `<!-- need a definition of frame -->` was added to the doc.

svgeesus:
Moves the definition of frame, which was introduced as part of `cLLi`, to the definitions section. This covers all three cases:

1. Static PNG (one frame)
2. Animated PNG, static image is first frame (all images in the PNG are frames)
3. Animated PNG, static image is not first frame (static image not counted as a frame)

References that definition from `cLLi`.

### DIFF
--- a/index.html
+++ b/index.html
@@ -413,8 +413,6 @@ with these exceptions:
       <!-- ************Page Break******************* -->
       <!-- ************Page Break******************* -->
 
-      <!-- need a definition of frame -->
-
       <dt><dfn>frame</dfn></dt>
 
       <dd>For static PNG, the <a>static image</a> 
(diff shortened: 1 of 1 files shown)


ProgramMax (approved):
LGTM.
Although on line 416, maybe remove the comment.

@palemieux are you happy with this being in the definitions section?

Issue #282, opened 2023-03-17
Privacy Considerations section should be expanded to cover risks deriving from cross-origin communication channels


AramZS:
Thanks for bringing this proposal to PING for review. This issue comes from the [requested review](https://github.com/w3cping/privacy-request/issues/105). 

While you requested specific review over the additions to the specification, I do see the Privacy Considerations section (currently placed in the decoders section) is new. I thought it would be appropriate to review the entire specification as it currently stands with that in mind. 

I did not find any particular privacy considerations that PNG or its new sections introduces intrinsic to the specification itself, however, the nature of PNGs are that they are capable and intended to be transmitted across contexts that the browser does not normally intend to communicate with each other. With this in mind, there are some risks to privacy that the PNG introduces that are not problems of the specification but just general dangers that I think adopters should be aware of to some extent. 

I think it would be appropriate to expand the Privacy Considerations section to mention these risks, to assure the spec is clear and make it more useful to anyone considering privacy risks or handling PNGs. This is not considered an urgent issue. This is my first encounter with the PNG specification and my review included an extensive reading and review of some linked materials associated with security risks and some raised issues, so please excuse any issues with my explanation here and let me know if these need additional clarity. Some of these stem from existing well-documented security issues, but have particular privacy implications that I think are worth documenting. 

Issues that I believe should be explored in the context of the Privacy Considerations section for completeness's sake that, from my reading, appear to be present:

- A mix of encoding and decoding techniques can be used to alter the size of images[^1] in ways that may not be anticipated by the top level domain that embeds an image. This can allow for DOM manipulation that can be used to encode information read by the top level domain or a third party embedded script. This may allow for communication of user data across third and first party contexts, or between two third parties across a top level page to which they would not normally have access or communication capabilities. 
  - Additionally the spec currently notes: `When a fatal condition occurs, the decoder should fail immediately, signal an error to the user if appropriate, and optionally continue displaying any [image data](https://www.w3.org/TR/2022/WD-png-3-20221025/#dfn-image-data) already visible to the user (i.e. "fail gracefully"). The application as a whole need not terminate.` If errors are added intentionally during the encoding stage, a forced early termination can be used to intentionally alter heights or widths in the DOM or be used to imply other user properties and be used as a fingerprinting vector. 

[...]

PR #306, opened 2023-04-27
Add known privacy issues, helps with #282

svgeesus:
Adds the three proposed paragraphs from https://github.com/w3c/PNG-spec/issues/282#issuecomment-1511449129 .

Does not close the issue, we might want to add encoder advice and we are also waiting for the original commenter to get back about our questions.


ProgramMax (approved):
LGTM

PR #316, opened 2023-06-02
Remove redundant note, fix #315

### LINKED ISSUE #315: Redundant note for fcTL
At the end of `fcTL` chunk section (also present in the original APNG spec):

> Note, at the end of the fcTL chunk is a 32-bit CRC checksum. The checksum is calculated using the fcTL chunk and includes the 'fcTL' chunk type field.

This is true for all chunks and is therefore redundant.

svgeesus:
Fixes https://github.com/w3c/PNG-spec/issues/315

### DIFF
--- a/index.html
+++ b/index.html
@@ -4922,10 +4922,7 @@ with these exceptions:
           step as long as the result is guaranteed to be identical. For example, if the default image is included in the animation,
           and uses a `blend_op` of `APNG_BLEND_OP_SOURCE`, clearing is not necessary because the entire output buffer will be
           overwritten.</p>
-          <!-- Not clear this para is needed, but the original specification had it -->
 
-          <p>Note, at the end of the <span class="chunk">fcTL</span> chunk is a 32-bit CRC checksum. The checksum is
-          calculated using the <span class="chunk">fcTL</span> chunk and includes the 'fcTL' chunk type field.</p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4922,10 +4922,7 @@ <h2><span class="chunk">fcTL</span> Frame Control Chunk</h2>
           step as long as the result is guaranteed to be identical. For example, if the default image is included in the animation,
           and uses a `blend_op` of `APNG_BLEND_OP_SOURCE`, clearing is not necessary because the entire output buffer will be
           overwritten.</p>
-          <!-- Not clear this para is needed, but the original specification had it -->
```
annevk, line 4925: Drive-by review: This comment still seems relevant. In particular "As noted earlier" followed by a normative requirement seems rather suspect. We shouldn't be requiring things twice. And the non-normative bits about optimizations might be better placed in a note.

svgeesus: The comment relates to the following paragraph, not the preceding one.

The PR deletes the needless duplicate requirement (which is true of all chunks, this one is not special) and also the now-irrelevant comment.

svgeesus:
@ProgramMax this simple PR has been hanging around for a while now, could you review please?

ProgramMax (approved):
LGTM

PR #318, opened 2023-06-07
Follow Encoding standard on UTF-8 encode and decode

### LINKED ISSUE #273: Text chunk processing should be in terms of the Encoding Standard
There's a couple of issues with the text here as far as I can tell:

* It's not clear how invalid UTF-8 byte sequences are handled.
* It's not clear if Latin1 means what it maps to in browsers (windows-1252) or isomorphic decode (https://infra.spec.whatwg.org/#isomorphic-decode).

I'm not sure it needs to say anything about encoders having data in other encodings. As there are no fields to store other encodings, it seems self-evident they have to convert.

svgeesus:
This PR only deals with UTF-8 encode and decode, for `iTXt`. It partially resolves #273

(We haven't yet got data to make a real Latin-1 vs 1252 decision for `tEXt` and `zTXt`.)

### DIFF
--- a/index.html
+++ b/index.html
@@ -5836,7 +5836,7 @@ scale_factor_Y = max(1.0, display_ratio/image_ratio)</code>
       "<code>\nnn</code>". See <a href="#13Security-considerations"></a>.</p>
 
       <p>When processing <a href="#11iTXt" class="chunk">iTXt</a> chunks,
-        decoders should follow <a href="https://encoding.spec.whatwg.org/#utf-8-decode">UTF-8 decode</a> in the [[[ENCODING]]].
+        decoders should follow <a href="https://encoding.spec.whatwg.org/#utf-8-decode">UTF-8 decode</a> in [[[ENCODING]]].
       </p>
 
       <p>Even though encoders are recommended to represent newlines as linefeed (hexadecimal 0A), it is recommended that decoders
--- a/index.html
+++ b/index.html
@@ -5499,7 +5499,7 @@ output = floor((input * MAXOUTSAMPLE / MAXINSAMPLE) + 0.5)
       </p>
 
       <p>When creating <a href="#11iTXt" class="chunk">iTXt</a> chunks,
-        encoders should follow <a href="https://encoding.spec.whatwg.org/#utf-8-encode">UTF-8 encode</a> in the [[[ENCODING]]].
+        encoders should follow <a href="https://encoding.spec.whatwg.org/#utf-8-encode">UTF-8 encode</a> in [[[ENCODING]]].
       </p>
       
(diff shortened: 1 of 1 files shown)


ProgramMax (approved):
LGTM

index.html
```diff
@@ -5494,8 +5494,15 @@ <h2>Text chunk processing</h2>
...
       (ISO 8859-1) character encoding, which limits the range of characters that can be used in these chunks. Encoders should
       prefer <a class="chunk" href="#11iTXt">iTXt</a> to <a class="chunk" href="#11tEXt">tEXt</a> and <a class="chunk" href=
-      "#11zTXt">zTXt</a> chunks. in order to allow a wide range of characters without data loss. Encoders must convert characters
-      that use local <a>legacy character encodings</a> to the appropriate encoding when storing text. Encoders should discourage
+      "#11zTXt">zTXt</a> chunks, in order to allow a wide range of characters without data loss. Encoders must convert characters
+      that use local <a>legacy character encodings</a> to the appropriate encoding when storing text. 
+      </p>
+
+      <p>When creating <a href="#11iTXt" class="chunk">iTXt</a> chunks,
+        encoders should follow <a href="https://encoding.spec.whatwg.org/#utf-8-encode">UTF-8 encode</a> in the [[[ENCODING]]].
```
annevk, line 5502:
```suggestion
        encoders should follow <a href="https://encoding.spec.whatwg.org/#utf-8-encode">UTF-8 encode</a> in [[[ENCODING]]].
```

index.html
```diff
@@ -5826,6 +5835,10 @@ <h2>Text chunk processing</h2>
       or <a class="chunk" href="#11zTXt">zTXt</a> chunk. Instead, they should be ignored or displayed in a visible notation such as
       "<code>\nnn</code>". See <a href="#13Security-considerations"></a>.</p>
 
+      <p>When processing <a href="#11iTXt" class="chunk">iTXt</a> chunks,
+        decoders should follow <a href="https://encoding.spec.whatwg.org/#utf-8-decode">UTF-8 decode</a> in the [[[ENCODING]]].
```
annevk, line 5839:
```suggestion
        decoders should follow <a href="https://encoding.spec.whatwg.org/#utf-8-decode">UTF-8 decode</a> in [[[ENCODING]]].
```

annevk (approved):
Do we have tests for BOM stripping? Might also be fun to see if you get the correct number of replacement code points.

ProgramMax:
I don't think we have tests for BOM stripping. I don't know a good way to automate that via WPT, since browsers (I am pretty sure?) don't expose that.

There aren't even many programs that read the text chunks which could be manually tested.

PR #321, opened 2023-07-10
Add note on unused trailing IDAT bytes

### LINKED ISSUE #158: Should unused trailing bytes be allowed in the data stream?
#96 mentions this. I'm pulling it out into a separate issue to keep separate issues independent.

It is surprisingly common for users of zlib to write out the entire buffer rather than just the used portion of the buffer. And unfortunately, many zlib users are sensitive to whether these unused trailing bytes are zeroed out or not.

Many PNG encoders & decoders run into this same issue (especially as many use zlib). There are many PNG images in the wild with unused trailing zeros in the compressed data stream.

Should we somehow specify that this is possible? Perhaps just a note for readers to be aware of this unintuitive detail.

ProgramMax:
This commit adds a note that unused trailing bytes may exist in the final IDAT chunk, explaining preferable handling.

Closes #158

### DIFF
--- a/index.html
+++ b/index.html
@@ -2986,6 +2986,8 @@ with these exceptions:
         <p>There may be multiple <span class="chunk">IDAT</span> chunks; if so, they shall appear consecutively with no other
         intervening chunks. The compressed datastream is then the concatenation of the contents of the data fields of all the
         <span class="chunk">IDAT</span> chunks.</p>
+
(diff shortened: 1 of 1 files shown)


ProgramMax:
I'm not sure on my wording here. Happy to take suggestions.

Also, IIRC a "note" carries some weight. So perhaps this shouldn't be a note. Thoughts?

svgeesus:
Testable implementation requirements should not be in a note

svgeesus (approved):
Looks good now

PR #325, opened 2023-07-18
Not a separate spec, fix #317

### LINKED ISSUE #317: APNG section reads like a separate spec
The APNG section talks about PNGs as if they are part of a separate spec. This makes sense, since the APNG wording comes from a separate spec.

Changing the wording to sound like it is all the same spec is not critical but would be nice.

svgeesus:
Hopefully this is clearer?

### DIFF
--- a/index.html
+++ b/index.html
@@ -1278,7 +1278,8 @@ with these exceptions:
       <section class="introductory">
         <h3>Introduction</h3>
 
-        <p>Animated PNG (APNG) is an extension of the PNG format, adding support for frame-based animated images. It is intended to
+        <p>Animated PNG (APNG) extends the original, static-only PNG format, 
+        adding support for <a>frame</a>-based animated images. It is intended to
         be a replacement for simple animated images that have traditionally used the GIF format [[GIF]], while adding support for
         24-bit images and 8-bit transparency, which GIF lacks.</p>
 


ProgramMax (approved):
LGTM

chrisn:
This change looks good. There are a couple of other places I noticed in this section that talk about the PNG spec as if it were a separate thing, that could possibly be rephrased:

* 4.9.1 has "as defined in previous versions of the PNG Specification" (do we still need to say this?)
* 4.9.2 has "since the PNG specification does not impose ordering restrictions on ancillary chunks" and describes sequence numbers in fcTL and fdAT chunks. To avoid potential duplication, could these be integrated into the relevant subsections of 11.3?

svgeesus:
> 4.9.1 has "as defined in previous versions of the PNG Specification" (do we still need to say this?)

I think so, because this is not new for the Web but is new in PNG Third Edition.

> 4.9.2 has "since the PNG specification does not impose ordering restrictions on ancillary chunks" and describes sequence numbers in fcTL and fdAT chunks. To avoid potential duplication, could these be integrated into the relevant subsections of 11.3?

Interesting point. However 4.9 is describing the _concept_ of a sequence number (so belongs in 4.x Concepts) , while this would need to be duplicated if explained inside `fcTL` and `fdAT`. So I feel the current ordering is better.

ProgramMax:
I agree with @chrisn on the 4.9.2 part of "since the PNG specification does not...". It sounds like it is talking in third person.
Although, it works. The alternate might be "since this specification does not...", which doesn't sound much better.
So I'm okay with it as-is.

svgeesus:
"this specification" does indeed read better

svgeesus:
See PR https://github.com/w3c/PNG-spec/pull/328

Issue #323, opened 2023-07-10
Call for Consensus: publish updated Working Draft


ProgramMax:
This GitHub issue is to establish consensus on publishing an updated Working Draft. W3C Working Group participants should reply with affirmation, dissent, or may abstain.

**Should we publish an updated Working Draft?**

#### Major changes in WD addressed

See [F.1 Changes since the [First Public Working Draft of 25 October 2022 (Third Edition)](https://www.w3.org/TR/2022/WD-png-3-20221025/).

Of note is the new cLLi and mDCv chunks, needed for tone mapping.

A new published Working Draft would put us in a good position for TAG review when the time comes.

Process
[Consensus is a core value to the W3C.](https://www.w3.org/2021/Process-20211102/#Consensus) I have assigned this issue to all [PNG WG participants who have their GitHub profile listed](https://www.w3.org/groups/wg/png/participants). I was unable to include Simon Thompson in the assignees list. Pinging @simontWork . Also, Said Abou-Hallawa does not have a GitHub account listed. I will try to email them.

A WG participant may register a [Formal Objection](https://www.w3.org/2021/Process-20211102/#FormalObjection) if they would like Director consideration.

WG participants may respond with a simple yes/no or similar. They may also elaborate. If dissenting, elaboration is encouraged in order to identify issues and find solutions. The [W3C process on managing dissent](https://www.w3.org/2021/Process-20211102/#managing-dissent) states that we will try to address concerns as much as is reasonable and should move on once that is done.

Call to action
Working group participants should reply with their stance on publishing an updated Working Draft.

svgeesus:
+1

palemieux:
+1

I recommend linking to issue #319 to alert readers on the ongoing discussion.

chrisn:
I also support publishing a new WD

svgeesus:
> A new published Working Draft would put us in a good position for TAG review when the time comes.

That time will be as soon as we publish the working draft :)

svgeesus:
Sadly, until we find a good volunteer or hire a Security expert, there is no security team to review specs. So we still ask, and give a few months in case some valiant member of the public steps up. Thus, in practice, all security reviews just time out.

svgeesus:
We need to go through the commit log and check that all substantive changes made are in fact reflected in the changes section, before publication.

ProgramMax:
+1 from me as well. That makes 6/10, simple majority reached.
[...]

PR #326, opened 2023-07-18
Changes from March to July, see #323

svgeesus:


ProgramMax (approved):
LGTM.
Out of curiosity, is there an importance threshold for things in this section? My mind was on "Clarifications don't warrant being called out". But thinking about it more, any change at all is worthy of being in patch notes.

svgeesus:
It is a judgement call how fine grained to go. In particular, markup changes and the like should probably not be listed. And where several edits were made to the same section, I combined them into one list item if it made sense to do so.

PR #329, opened 2023-07-20
Consistently use example and note classes fix #322

### LINKED ISSUE #322: Some NOTEs and EXAMPLEs aren't rendering properly
[8.1 Interlace methods](https://w3c.github.io/PNG-spec/#8InterlaceMethods) has a "NOTE" at the bottom. It is marked up with class="Note". Notice it is a capital "Note". Elsewhere, there is class="note" (lowercase) which behaves correctly.

There are other class="Note"s as well.

[5.4 Chunk naming conventions](https://w3c.github.io/PNG-spec/#5Chunk-naming-conventions) has an "EXAMPLE" which is not marked up with class="example". The same is in [11.3.2.1 cHRM Primary chromaticities and white point](https://w3c.github.io/PNG-spec/#11cHRM) and elsewhere, too.

svgeesus:


ProgramMax (approved):
LGTM

PR #333, opened 2023-07-25
Added Dark mode in index page as well as extentions page

ankit071105:
# added light mode
Screenshot 2023-07-25 at 11 18 54 PM

#  added Dark mode 
Screenshot 2023-07-25 at 11 19 00 PM

### DIFF
--- a/REC-PNG-20031110-errata.html
+++ b/REC-PNG-20031110-errata.html
@@ -9,7 +9,102 @@
         </style>
         <link rel="stylesheet" type="text/css" href="http://www.w3.org/StyleSheets/TR/base" />
         </head>
+        <style>
+               :root {
+    --bg: rgb(137, 144, 158);
+    --bg-input: #141419;
+}
+
+
+:root{
+    --bg: linear-gradient(to left, #effafd 0%, #f3e8f9 100%);
+   
+    --bg-input: #0d0d14;
+}
+
+
+/* light mode edited */
+
+.light-mode {
+    --bg: linear-gradient(to left, #2a2d30, #080a0c);
+    --bg-input: #fcf9f9;
+}
+
+body {
+    background: var(--bg);
+    color: var(--bg-input)
+    /*font-weight: 500;
+    display: flex;*/
+}
+
+.switch {
+    position: absolute;
+    margin-left: -9999px;
+    visibility: hidden;
+}
+
+.switch+label {
+    display: block;
+    position: relative;
+    cursor: pointer;
+    outline: none;
+    user-select: none;
+}
+
+.switch+label {
+    padding: 2px;
+    width: 52px;
+    height: 27px;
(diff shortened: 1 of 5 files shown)


ProgramMax:
In general, I like the idea of adding dark mode support.
I have a few general thoughts but we might not want to act on them just yet.
For example, we should probably use prefers-color-scheme to respect the user's default. The toggle switch is then not required (but maybe we want to leave it anyway). And there are probably better options than using !important. Plus the toggle shouldn't have its own section, which gets added to the table of contents.

But the reason we shouldn't act on these yet is:
This is styled to match existing W3C designs and other specs. We don't want to break that without W3C designer approval. @svgeesus do you know any designers in W3C we should talk to? Presumably, we would want to add dark mode to all the specs, not just this one. Talking to the designers will help us address the issues I mentioned above.

That said, I am happy for the PNG spec to be the test subject in a more broad add-dark-mode-everywhere effort.

For now, lets wait to get in contact with some W3C designers.

svgeesus:
I discussed this with the W3C webmasters. It turns out that there **is** official dark mode support, since 2 years ago; it adds a toggle onto the page as well. However it has not been [well documented](https://www.w3.org/StyleSheets/TR/2016/README.html#dark) and not widely advertised because it is unsatisfactory in some ways:

1. For ReSpec, the stylesheet needs to be added manually after exporting the generated ReSpec view.  There is no automatic support in ReSpec yet although there has been [a PR to work on this](https://github.com/w3c/respec/pull/3236)
2. For BikeShed, there is no automatic support although there have been [requests for it](https://github.com/speced/bikeshed/issues?q=is%3Aissue+is%3Aopen+dark).
3. Automatically enabling it cannot be done because many specs have figures which were not designed with dark mode in mind and they become unreadable. Same for custom styling that some specs use.

Webmasters are currently trying to revive those efforts so that the stylesheet can be officially supported.

The SVG figures in the PNG spec would need to have dark mode styling added and the main document toggle would need to toggle those graphics as well.

Meanwhile, since there is an official solution in the works, I'm closing this PR.

svgeesus:
@ankit071105 Thanks for taking the trouble to work on this, the effort is appreciated, but we are better going with a W3C-wide change for this.

The changes you made would also result in the documents failing the W3C publications checker, which is quite strict.

svgeesus:
Here is a [staged version of the PNG spec with the dark-mode stylesheet added](https://mir.w3.org/~denis/png.html). Some issues:

- All examples are too low contrast, needs alternate dark-mode styling
- Many graphics loose black or dark arrows, boxes etc. Not too much work to create alternate dark-mode versions, but the switcher would need to change image urls too (I have asked if that could be considered)
- Styling of chunk names needs to be changed for dark mode

Interestingly a couple of ASCII-art graphics or fake tables, which I thought we had got rid of, need to be fixed in both light mode and dark mode.

PR #352, opened 2023-08-31
Added changes since WD of 20 July 2023

svgeesus:
We were missing changes since the [20 July 2023 WD](https://www.w3.org/TR/2023/WD-png-3-20230720/), and the CR transition needs a link to the latest changes


ProgramMax (approved):
LGTM

PR #363, opened 2023-11-01
Zero-length chunk data, fix #356

### LINKED ISSUE #356: Zero-length `IDAT` (and `fdAT`)
I came across this comment while attempting to read the source of `pngcheck` (not recommended):

>       /* We just want to check that we have read at least the minimum (10)
>       * IDAT bytes possible, but avoid any overflow for short ints.  We
>       * must also take into account that 0-length IDAT chunks are legal.
>       */

I went looking for spec text to back this up, and there isn't any (either explicitly allowing it, or specifically dis-allowing). It has to be inferred:

> There may be multiple IDAT chunks; if so, they shall appear consecutively with no other intervening chunks. The compressed datastream is then the concatenation of the contents of the data fields of all the IDAT chunks.
([1.2.3 IDAT Image data]())

and
> Similarly, there is no required correlation between the structure of the [image data](https://www.w3.org/TR/png-3/#dfn-image-data) (i.e., scanline boundaries) and [deflate](https://www.w3.org/TR/png-3/#dfn-deflate) block boundaries or [IDAT](https://www.w3.org/TR/png-3/#11IDAT) chunk boundaries. The complete filtered PNG image is represented by a single [zlib](https://www.w3.org/TR/png-3/#dfn-zlib) datastream that is stored in a number of [IDAT](https://www.w3.org/TR/png-3/#11IDAT) chunks.
([10.2 Compression of the sequence of filtered scanlines](https://www.w3.org/TR/png-3/#10CompressionFSL))

Given that `pngcheck` explicitly allows `IDAT` with zero-length data, and the 2007 [APNG tests](https://philip.html5.org/tests/apng/tests.html) from Philip Taylor includes explicit tests for zero-length `IDAT` and zero-length `fdAT', which browsers pass, then it seems that the spec would be improved by mentioning this consequence explicitly.

svgeesus:
Explicitly remind that concatenated data from IDAT and fdAT may include zero-length chunks

### DIFF
--- a/index.html
+++ b/index.html
@@ -1680,7 +1680,8 @@ with these exceptions:
 
         <tr>
           <td>Chunk Data</td>
-          <td>The data bytes appropriate to the chunk type, if any. This field can be of zero length.</td>
+          <td>The data bytes appropriate to the chunk type, if any. 
+            <span id="zero-length-data">This field can be of zero length.</span></td>
         </tr>
 
         <tr>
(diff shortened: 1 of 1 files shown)


svgeesus:
@ProgramMax

ProgramMax (approved):
LGTM

PR #373, opened 2023-11-09
Improve wording in cICP section

chrisn:
"Currently only RGB is the supported color model in PNG" was awkward phrasing.


svgeesus (approved):
Yes, reads better

ProgramMax (approved):
LGTM

Issue #369, opened 2023-11-09
Does mDCV require cICP?


simontWork:
In the mDCV description, the following sentence suggests that an accompanying cICP is required:

"Color Primaries and White Point characteristics can be derived from cICP chunk formats."

If this is so, do we need to mandate cICP is used if mDCV is present?

digitaltvguy:
I've added new text for this requiring cICp if mDCv is present.  I will create a pull request after reviewing all other issues related to cICp and mDCv

svgeesus:
@digitaltvguy looks like you need to rebase so that your fork catches up to this repository

digitaltvguy:
fixed (i hope).  See pull request "Added conditions for use of mDCv which require cICp chunk #378"

digitaltvguy:
I think C.B. helped resolve this. My future updates will come via branches with pull request.

PR #375, opened 2023-11-10
Added language to address Issue #369

digitaltvguy:
• Is cICP needed when mDCv is used
• Added values in table 15 which were missing (color primaries as CIE 1931; luminance as cd/m2)

### DIFF
--- a/index.html
+++ b/index.html
@@ -3676,7 +3676,7 @@ with these exceptions:
           <p>Each of the fields of the <span class="chunk">cICP</span> chunk corresponds to the parameter of the same name in
           [[ITU-T-H.273]].</p>
 
-          <p>Currently only RGB is the supported color model in PNG, and as such <code>Matrix Coefficients</code> shall be set to <code>0</code>.</p>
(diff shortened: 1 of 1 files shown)


svgeesus:
@digitaltvguy it seems your fork is several months behind, could you rebase and then apply your edits so the PR only includes the relevant changes?

ProgramMax:
@svgeesus is correct here. I may have done you a disservice by recommending you create branches on your fork. Because now you need to also make sure your fork is up to date before you make the branch.

Your fork is here https://github.com/digitaltvguy/PNG-spec/ -- On that page, above the list of files will be a button to update. I forget the exact text of the button.

Updating the fork will require resolving any conflicts that may have been introduced. There are two approaches for this: rebasing and merging. What is what @svgeesus meant when he said you'll need to rebase.

If your changes are small and you can recreate them, it might actually be easier to abandon your changes, update, and make a new branch. That way you don't need to deal with conflicts.

I'm happy to get on a call with you some time and walk you through these if you want.

PR #381, opened 2023-12-04
Clarify sequence numbers and zero-length fdAT, would fix #379

### LINKED ISSUE #379: Clarify that a frame with multiple fdAT's has only one sequence number
The current spec isn't very clear whether the fdAT chunks contain only one sequence number per-frame or if every fdAT starts with a sequence number. The most clear indication is probably the confirmation that [zero length fdAT chunks are allowed](https://github.com/w3c/PNG-spec/issues/356), which is incompatible requiring all fdAT's to start with a sequence number.

The most confusing spot is probably Table 27. It seems to show the structure of a single fdAT chunk consisting of a 4-byte sequence number followed by n-bytes of frame data. However, my understanding is that structure is actually the breakdown _after concatenating all the fdAT's for the frame_:

> fdAT chunk contents:
> Name | Size
> -- | --
> sequence_number | 4 bytes
> frame_data | n bytes

The subsequent text talks about the concatenation of multiple fdAT chunks, but completely omits any mention sequence numbers:

> The compressed datastream is then the concatenation of the contents of the data fields of all the fdAT chunks within a frame          (noting that data fields may be of zero length).

svgeesus:
No longer claims that `fdAT` can be zero length, but clarify the `frame_data` can be zero length.

Generally tighten wording around chunk concatenation, explicitly mention "in ascending sequence number order".

Link back to the section in Concepts on sequence numbers.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4943,7 +4943,7 @@ with these exceptions:
             </tr>
           </table>
 
-          <p>`sequence_number` defines the sequence number of the animation chunk, starting from 0. It is encoded as a <a>PNG
+          <p>`sequence_number` defines the <a href="#4Concepts.APNGSequence">sequence number</a> of the animation chunk, starting from 0. It is encoded as a <a>PNG
           four-byte unsigned integer</a>.</p>
 
           <p>`width` and `height` define the width and height of the following frame. They are encoded as <a>PNG four-byte unsigned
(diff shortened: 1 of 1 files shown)


ProgramMax (approved):
LGTM

PR #402, opened 2024-01-05
Link to normative requirement, fix #361

### LINKED ISSUE #361: `tRNS` normative text in note
in [`tRNS`](https://w3c.github.io/PNG-spec/#11tRNS) there is a Note (thus, non-normative) containing a normative requirement:

> NOTE For 16-bit [greyscale](https://w3c.github.io/PNG-spec/#dfn-greyscale) or [truecolour](https://w3c.github.io/PNG-spec/#3truecolour) data, only pixels matching the entire 16-bit values in tRNS chunks are transparent. Decoders have to postpone any sample depth rescaling until after the pixels have been tested for transparency.

"Have to" sounds like it should be "must".

Simple fix is to remove the "Note".

svgeesus:
From the note, link to the actual normative requirement for 16-bit compare

### DIFF
--- a/index.html
+++ b/index.html
@@ -3191,7 +3191,9 @@ with these exceptions:
           <p>A <span class="chunk">tRNS</span> chunk shall not appear for <a>colour types</a> 4 and 6, since a full alpha channel
           is already present in those cases.</p>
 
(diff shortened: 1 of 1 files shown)


ProgramMax (approved):
LGTM

Issue #374, opened 2023-11-09
Inconsistent colour / color


chrisn:
A minor issue, sorry. Should we pick one and use it throughout the spec?

svgeesus:
The [original PNG spec from 1996](https://www.w3.org/TR/REC-png-961001) used color. see for example [2.2 Color values](https://www.w3.org/TR/REC-png-961001#DR.Color-values).

The reason being that the [W3C style guide](https://www.w3.org/Guide/manual-of-style/#Spelling) says, for spelling

> W3C uses U.S. English (e.g., "standardise" should read "standardize" and "behaviour" should read "behavior").

[PNG Second Edition in 2003](https://www.w3.org/TR/2003/REC-PNG-20031110/) used colour, see for example [4.2 Colour spaces](https://www.w3.org/TR/2003/REC-PNG-20031110/#4Concepts.ColourSpaces). 

The reason being that it was also published with identical content as ISO/IEC 15948:2003 (E) and ISO style rules require the use of International English (i.e. English without [Webster's US-specific spelling reform of 1828](https://www.quora.com/Why-do-a-large-number-of-words-in-the-Americans-version-of-the-English-language-differ-in-spelling-and-or-pronunciation))

( I remember when Lilley et. al _Colour in Computer Graphics_ was published in the UK in 1993 and made available online, I rapidly got a comment that "You have mis-spelled color, not just once, but throughout! Did you not think to use a spell checker?")

Of course whichever is chosen there will still be inconsistencies - If we use International English throughout, we will still refer to the International Color Consortium for example.

But I agree we should be consistent where possible and my preference would be to follow the W3C style guide and all other W3C specs and use the US English spelling.

Note that the links embed the spelling choices too:

https://www.w3.org/TR/REC-png-961001#DR.Color-values
https://www.w3.org/TR/2003/REC-PNG-20031110/#4Concepts.ColourSpaces

Given that the First Edition was in force for 1996-2003 and the Second Edition from 2003-2023 we should preserve link compatibility on fragment identifiers with the Second Edition, which we have been doing.

https://www.w3.org/TR/png-3/#4Concepts.ColourSpaces

ProgramMax:
I feel like early in Third Edition development I read something (likely W3C?) saying to use international English. Maybe I'm misremembering.

In either case, I agree we should be consistent without breaking inbound links.

svgeesus:
PR to use color in visible text, while preserving colour on inbound links.

[...]

PR #411, opened 2024-01-13
Use US-English color in visible text, preserving incoming links that …

svgeesus:
…use colour #374

### DIFF
--- a/index.html
+++ b/index.html
@@ -316,17 +316,17 @@ needs update, comment out for now
   <section id="sotd">
     <p>This specification is intended to become an International Standard, but is not yet one. It is inappropriate to refer to this
     specification as an International Standard.</p>
-    <p>The following features are at-risk and may be removed from this specification 
+    <p>The following features are at-risk and may be removed from this specification
       if there is insufficient implementation support:</p>
     <ul>
       <li>The <a class="chunk" href="#eXIf">eXIf</a> chunk</li>
     </ul>
-    <p>“At-risk” is a W3C Process term-of-art, and does not necessarily imply 
-      that the feature is in danger of being dropped or delayed. 
-      It means that the WG believes the feature may have difficulty being 
-      interoperably implemented in a timely manner, and marking it as such 
-      allows the WG to drop the feature if necessary when transitioning to the 
-      Proposed Recommendation stage, without having to first publish a new 
+    <p>“At-risk” is a W3C Process term-of-art, and does not necessarily imply
+      that the feature is in danger of being dropped or delayed.
+      It means that the WG believes the feature may have difficulty being
+      interoperably implemented in a timely manner, and marking it as such
+      allows the WG to drop the feature if necessary when transitioning to the
+      Proposed Recommendation stage, without having to first publish a new
       Candidate Recommendation without the feature.</p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -843,26 +843,26 @@ <h2 id="4Concepts.Sourceimage">Images</h2>
     <section id="4Concepts.ColourSpaces">
       <h2>Colour spaces</h2>
```
annevk, line 844: Seems like this should be renamed. (Not the ID of course.)

index.html
```diff
@@ -1021,14 +1021,14 @@ <h2>PNG image</h2>
...
-        <dt><dfn>Truecolour with alpha</dfn></dt>
+        <dt><dfn>Truecolor with alpha</dfn></dt>
         <dd>Each <a>pixel</a> consists of four <a>samples</a>: red, green, blue and <a>alpha</a>.</dd>
 
         <dt><dfn>Greyscale with alpha</dfn></dt>
         <dd>Each <a>pixel</a> consists of two <a>samples</a>: a <a>grey sample</a> and an <a>alpha</a> sample.</dd>
 
         <!-- Maintain a fragment named "3truecolour" to preserve incoming links to it -->
-        <dt><dfn id="3truecolour">Truecolour</dfn></dt>
+        <dt id="truecolor"><dfn id="3truecolour">Truecolor</dfn></dt>
```
annevk, line 1031: I would either not add an ID or give the `<dfn>` the improved ID.

index.html
```diff
@@ -5618,17 +5618,17 @@ <h2>Interlacing</h2>
     <section id="12Filter-selection">
       <h2>Filter selection</h2>
 
-      <p>For images of <a>colour type</a> 3 (indexed-colour), filter type 0 (None) is usually the most effective. Colour images
-      with 256 or fewer colours should almost always be stored in <a>indexed-colour</a> format; <a>truecolour</a> format is likely to be
+      <p>For images of <a>color type</a> 3 (indexed-color), filter type 0 (None) is usually the most effective. Colour images
```
annevk, line 5621:
```suggestion
      <p>For images of <a>color type</a> 3 (indexed-color), filter type 0 (None) is usually the most effective. Color images
```
Maybe do another search that's case-insensitive.

annevk (commented):
Thanks! Style guides FTW. (Not approving as I'm not a reviewer.)

index.html
svgeesus: Good catch! I see that I had "match case" on in my search-and-replace, so I just did another pass changing Colour to Color as needed.

index.html
svgeesus: We want to preserve inbound link fragments where possible, to keep compatibility from the 2003 second edition. That is why this one has both; it enables ReSpec `<a>truecolor</a>` to work while also preserving the older id as well.

annevk: ReSpec doesn't key off the `dfn`? Weird but okay.

svgeesus (commented):
Thanks for those, good to have consistent usage

svgeesus:
@ProgramMax does this look ok to you?

chrisn (approved):

PR #416, opened 2024-01-24
Rephrase cLLi chunk description

chrisn:
This PR rephrases the cLLI chunk description. The current text has:

> If present, the cLLi chunk identifies two characteristics of HDR content:

which seems intended to be followed directly by the description of MaxFALL and MaxCLL. This PR reorders the paragraphs (with some slight rewording) to do that.

### DIFF
--- a/index.html
+++ b/index.html
@@ -544,7 +544,7 @@ with these exceptions:
       </dt>
 
       <dd>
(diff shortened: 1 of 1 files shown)


svgeesus (approved):
LGTM

PR #421, opened 2024-02-20
Space not allowed in bibref

svgeesus:
The HDR10 reference was failing because the bibliography entry called it HDR 10 (with a space) so it didn't match. Plus, spaces are not allowed there. So I fixed it by using HDR10.


svgeesus:
Sorry, this was my bad originally.

chrisn (approved):

ProgramMax (commented):
post-mortem LGTM

PR #422, opened 2024-02-21
Updated list of changes, to 20 Feb 2024

svgeesus:
The list of changes since the [21 Sept 2023 CR Snapshot](https://www.w3.org/TR/2023/CR-png-3-20230921/) was missing, so I added it from the [GitHub commit history](https://github.com/w3c/PNG-spec/commits/main/index.html?before=0ef99d495a6e9c25a404de2c53b5a43b8d7cd26d+35).

I noticed that some changes before that were missing too, so added those as well.

I omitted changes that we made, and then backed out without publishing to /TR in between (such as putting `eXIf` at-risk and then no longer at-risk)

### DIFF
--- a/index.html
+++ b/index.html
@@ -7529,10 +7529,47 @@ will need to be replaced by copies of lines 17-23, but processing background ins
   <section id="F-ChangeList" class="appendix informative">
     <h2 class="Annex">Changes</h2>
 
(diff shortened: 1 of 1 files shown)


chrisn (approved):
Thanks @svgeesus!

PR #434, opened 2024-03-01
Recommend using extended transfer functions to handle negative values

### LINKED ISSUE #420: Reccomend how to deal with negative values
Following a long discussion on the mailing list, perhaps we should add a note somewhere in the text that when the FullRange flag is set to zero, the recommended practice is

out = sign(in) * FUNCTION(abs(in))

Where function can be for example EOTF, OETF^-1

svgeesus:
Would fix #420

### DIFF
--- a/index.html
+++ b/index.html
@@ -711,6 +711,12 @@ with these exceptions:
       <dd>vacuum tube containing one or more electron guns, which emit electron beams that are manipulated to display images on a
       phosphorescent screen.</dd>
 
+      <dt>Electro-Optical Transfer Function</dt>
+      <dt>
+        <abbr title="Electro-Optical Transfer Function">EOTF</abbr>
+      </dt>
+      <dd>The <a>transfer function</a> between the electrical or digital domain and light energy. It defines the amount of light emitted by a display for a given input signal.</dd>
+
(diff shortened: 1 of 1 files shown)


simontWork (approved):
Looks good to me. (I've just realised I made a pull request against my fork rather than W3C's with similar text)

ProgramMax (approved):
LGTM

index.html
```diff
@@ -3725,6 +3725,17 @@ <h2><span class="chunk">cICP</span> Coding-independent code points for video sig
               PNG format, it should be noted that this format is not part of the [[ITU-R-BT.709]] standard.
           </aside>
 
+          <p>If <code>Video Full Range Flag</code> is <code>0</code>
+            (a <a>narrow-range image</a>), recommended practice 
+            is to define transfer functions 
+            such as EOTF or inverse OETF
```
fintelia, line 3731: The EOTF acronym doesn't seem to be defined anywhere else in the document. It would be nice to avoid introducing new terminology just to explain how to handle the Full Range flag, but at very least the terminology should be defined

svgeesus: Better?

svgeesus:
Re-requesting reviews due to addition of EOTF and OETF definitions

PR #439, opened 2024-03-19
437 color metadata should be before image data

svgeesus:


chrisn (changes requested):
The substance here looks fine, but the HTML formatting needs fixing.

chrisn:
> The substance here looks fine, but the HTML formatting needs fixing.

FIxed in https://github.com/w3c/PNG-spec/pull/439/commits/32d0cc3389fd1f54758a1bfbcd4b3d737adee2e0.

chrisn (approved):

PR #451, opened 2024-06-04
Add private bit depth and colour type fields

### LINKED ISSUE #446: PNGv2 12.11.2 -> PNGv3 12.10.2 transcription issue
v2 12.11.2:

> 12.11.2 Private type and method codes
> This specification defines the meaning of only some of the possible values of some fields. For example, only compression method 0 and filter types 0 through 4 are defined in this International Standard. Numbers greater than 127 shall be used when inventing experimental or private definitions of values for any of these fields. Numbers below 128 are reserved for possible public extensions of this specification through future standardization (see 4.9 [Extension and registration](https://www.w3.org/TR/2003/REC-PNG-20031110/#4Concepts.Registration)). The use of private type codes may render a datastream unreadable by standard decoders. Such codes are strongly discouraged except for experimental purposes, and should not appear in publicly available software or datastreams.

Becomes v12.10.2:

>12.10.2 Use of non-reserved field values
>Encoders MAY use non-reserved field values for experimental or private use.

I feel this is excessive, but that's not an issue.  Rather the issue is the extra specification:

> 5.8 Private field values
> Values greater than or equal to 128 in the following fields are private field values:
> 
> [compression method](https://www.w3.org/TR/png-3/#10CompressionCM0)
> [interlace method](https://www.w3.org/TR/png-3/#8InterlaceMethods)
> [filter method](https://www.w3.org/TR/png-3/#9FtIntro)
> These [private field values](https://www.w3.org/TR/png-3/#dfn-private-field-values) are neither defined nor reserved by this specification.
> 
> [Private field values](https://www.w3.org/TR/png-3/#dfn-private-field-values) MAY be used for experimental or private semantics.
> 
> [Private field values](https://www.w3.org/TR/png-3/#dfn-private-field-values) SHOULD NOT appear in publicly available software or datastreams since they can result in datastreams that are unreadable by PNG decoders as detailed at [13. PNG decoders and viewers](https://www.w3.org/TR/png-3/#13Decoders).

So the bug of consequence is that this disallows private "color types" which were explicitly allowed in PNGv2; section 5.8 is new.

It also allows private big depth, which was not allowed before, but also it allows addition of compression method, filter method, filter type and interlace method without using the private bit flag (bit 7).  The bit flag is there to make it clear to everyone that this is a private, non-approved, extension/change.

ProgramMax:
In Second Edition, all fields were allowed to have private values when the value was 128 or greater.

This shouldn't have been *all* fields, since some rely on values greater than 128 such as height and width.

Third Edition explicitly listed fields which could be private. Currently, it lists 3 of the 5 IHDR fields (ignoring width and height). However, those other two fields would be useful as private, too. Those fields are bit depth and colour type.

This commit adds those two fields to the list of explicitly allowed private fields.

Closes #446

### DIFF
--- a/index.html
+++ b/index.html
@@ -2311,6 +2311,12 @@ with these exceptions:
 
       <ul>
         <li>
+          bit depth
+        </li>
+        <li>
+          <a href="#3colourType">colour type</a>
+        </li>
+        <li>
           <a href="#10CompressionCM0">compression method</a>
         </li>
 


ProgramMax:
@jbowler Here is the patch for your suggestion. Thank you for catching that.

The "bit depth" field intentionally does not link anywhere. Elsewhere in the doc, we talk about bit depth without links. There is no section specifically defining it or an entry in the definitions.

svgeesus (approved):
Nice catch

PR #454, opened 2024-06-21
WPT now uses  macOS 14.5 so live tests pass

svgeesus:
The implementation report used to have special pleading that some tests in Safari would pass, if only wpt was running macOS 14 instead of 13.6. This is no longer needed because WPT is now running macOS 14.5.

Also updated browser versions to current WPT ones.

Also fixed an error (one failing test was incorrectly shown in the table as the preceding test, my bad)

### DIFF
--- a/Implementation_Report_3e/index.html
+++ b/Implementation_Report_3e/index.html
@@ -48,7 +48,7 @@
                 <dt>Editor:</dt>
                 <dd>Chris Lilley</dd>
                 <dt>Last modified:</dt>
-                <dd>2024-03-28</dd>
+                <dd>2024-06-21</dd>
             </dl>
         </div>
         <main>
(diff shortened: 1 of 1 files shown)


Implementation_Report_3e/index.html
```diff
@@ -199,17 +199,17 @@ <h3>Frame control tests</h3>
...
-                            <th>Safari TP 190</th>
+                            <th>Safari TP 196</th>
                         </tr>
                         <tr class="test">
                             <td class="testname">fcTL-acTL-ordering.html </td>
                             <td class="result passes-all"></td>
                             <td class="result passes-all"></td>
                             <td class="result passes-all"></td>
-                            <td class="result passes-all"></td>
+                            <td class="result passes-none"></td>
```
ProgramMax, line 212: Just checking that this change was intentional.
Did it regress? Is it a correction of a previous mistake?

ProgramMax (approved):
LGTM but one possible mistake to check

Implementation_Report_3e/index.html
svgeesus: "Also fixed an error (one failing test was incorrectly shown in the table as the preceding test, my bad)

PR #457, opened 2024-07-07
cICP->Matrix Coefficients is always 0, not 1

Aemony:
Third byte always has the value of `0`, not `1`, according to https://w3c.github.io/PNG-spec/#cICP-chunk


ProgramMax (approved):
Good catch. This is correct. I checked it against the spec and H.273.

We just took a snapshot of the spec so I was worried this mistake might require a new snapshot. But luckily, this mistake is in the explainer and not the spec. Phew.

PR #460, opened 2024-07-16
2024 07 18 crd

svgeesus:
Changes for publication as a Candidate Recommendation Draft.

Notice that we [did **not** get permission](https://github.com/w3c/transitions/issues/636#issuecomment-2223346229) for a Candidate Recommendation Snapshot, and thus need to go through Horizontal Review again, _plus_ document review from [external organizations as specified in the charter](https://www.w3.org/Graphics/PNG/png-2023.html#coordination). But to restart horizontal review we first need a CRD, which will hopefully be published this Thursday.

There was also concern we didn't get review for substantive [changes since FPWD](https://www.w3.org/TR/png-3/#changes-20230720). The sole substantive change was:

 - Mandated current browser handling of out-of-range palette indices

which we [have a test for]() as [documented in the implementation report](https://w3c.github.io/png/Implementation_Report_3e/#palette) and that test is passed in Chrome, Edge and Safari (so, 2 independent implementations) and is, as we said, current browser behavior. Only Firefox needs to update to pass, which they [said they would do](https://bugzilla.mozilla.org/show_bug.cgi?id=1862235). So I think that criticism is baseless.

Still, here are the changes needed to publish the CR Draft and we can go from there.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3,28 +3,25 @@
 <head>
   <title>Portable Network Graphics (PNG) Specification (Third Edition)</title>
   <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
-  <!-- <link rel="stylesheet" href="./isostyle.css" type="text/css"> -->
 
   <style type="text/css">
-    /* remove annoying green colour from definition terms */
-    dt {color: black}
     span.chunk { color: #622;}
   </style>
   <script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove" defer></script>
   <script class="remove">
     var respecConfig = {
       group: "png",
-      specStatus: "CR",
-      crEnd: "2023-12-21",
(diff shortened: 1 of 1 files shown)


chrisn (approved):

index.html
```diff
@@ -1419,7 +1413,7 @@ <h3>Introduction</h3>
         24-bit images and 8-bit transparency, which GIF lacks.</p>
 
         <p>APNG is backwards-compatible with earlier versions of PNG; a non-animated PNG decoder will ignore the ancillary
-        APNG-specific chunks and display the <a>static image</a>.</p>
+        Apngific chunks and display the <a>static image</a>.</p>
```
ProgramMax, line 1416: I think this change wasn't intentional.

ProgramMax (approved):
LGTM except one potential accidental change

PR #465, opened 2024-09-15
moved cd/m2 to name column

digitaltvguy:
This moves cd/m2 from divisor column and moves it to the name column

### DIFF
--- a/index.html
+++ b/index.html
@@ -3904,13 +3904,13 @@ with these exceptions:
             </tr>
 
             <tr>
-              <td>Mastering display maximum luminance</td>
+              <td>Mastering display maximum luminance (measured in cd/m<sup>2</sup>)</td>
               <td>4 bytes</td>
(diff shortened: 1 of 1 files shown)


svgeesus (approved):

index.html
```diff
@@ -3904,13 +3904,13 @@ <h2><span class="chunk">mDCv</span> Mastering Display Color Volume</h2>
             </tr>
 
             <tr>
-              <td>Mastering display maximum luminance</td>
+              <td>Mastering display maximum luminance (measured in cd/m<sup>2</sup>)</td>
               <td>4 bytes</td>
-              <td>0.0001 cd/m<sup>2</sup></td>
```
svgeesus, line 3909: why is this removed here, but not in min luminance?

ProgramMax (commented):
post-facto LGTM

PR #472, opened 2024-09-30
Added table with 16bit full and narrow range values

### LINKED ISSUE #312: Narrow range and `cICP`
See [later discussion on narrow range and full range](https://github.com/w3c/png/issues/312#issuecomment-2325281113)

===========

This can be tested in WPT by converting full-range sRGB values to one of the common encodings (such as 2100 PQ, 2100 HLG, Display P3, linear sRGB) and making a PNG with those values and the appropriate `cICP` chunk. A pass displays the same colors as an sRGB reference image.

For non-browser support, detecting and correctly decoding the chunk demonstrates support.

digitaltvguy:
to replace current definition for "reference image" which only described Full Range Quantization.

### DIFF
--- a/index.html
+++ b/index.html
@@ -780,9 +780,43 @@ with these exceptions:
         <!-- Maintain a fragment named "3channel" to preserve incoming links to it -->
         <dfn id="3channel">channel</dfn>. Each channel has a sample depth in the range 1 to 16, which is the number of bits used by every sample in the
         channel. Different channels may have different sample depths. The red, green, and blue samples determine the intensities of
-        the red, green, and blue components of the pixel's color; if they are all zero, the pixel is black, and if they all have
+        the red, green, and blue components of the pixel's color. 
+        
+        <p>The following table describes the quantization of image black to image peak white (signal range). By default the signal range is full-range 
+          but it can also be represented as narrrow range, which conforms to quantization in broadcast video which typically uses the YCbCr Color 
+          Representation.  cICP can specify this specific range:</p>
+
+          <br>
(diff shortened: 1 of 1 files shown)


svgeesus (changes requested):
I think this proposed change needs Working Group discussion before being merged.

- strange mention of YCbCr, which PNG does not support
- replaces 2^sampledepth with 16

@ProgramMax

index.html
```diff
@@ -780,9 +780,43 @@ <h2 id="4Concepts.Sourceimage">Images</h2>
...
+          </caption>
+
+         
+          <tr>
+            <th>16bit Signal Range (Quantization for Full and Narrow Ranges)</th>
+            <th>Black</th>
+            <th>White</th>
+          </tr>
+
+          <table border="1">
```
ProgramMax, line 802: I think this line can be removed. It accidentally adds an extra table inside the other table. Notice, it also never gets closed.

ProgramMax (commented):
Agreed with CL.
I added it to the discussion topics.
We should also mark it as informative.

ProgramMax:
Also, I notice this is in § 4.2 Images - Reference image. Perhaps that section should instead be changed to say something like "1 to 16 bits, although not all bits might be used. The sBIT chunk (link) confines the bit range, and the cICP chunk (link) might declare the image is narrow range."

Then the cICP chunk could have these example values instead of 4.2

jbowler:
Protocol: close pull requests that have both been mooted and the cancelled...

PR #473, opened 2024-10-30
Rename mDCv to mDCV

ProgramMax:
The mDCV chunk depends on image data. As a result, it is not safe to copy if that pixel data changes.

This commit renames the mDCv chunk to mDCV to reflect the unsafe-to-copy behavior.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3838,7 +3838,7 @@ with these exceptions:
           <p>The four-byte chunk type field contains the hexadecimal values</p>
 
           <pre>
-<!-- 109 68 67 118 -->6D 44 43 76
+<!-- 109 68 67 86 -->6D 44 43 56
 </pre>
 
           <p>If present, the <span class="chunk">mDCV</span> chunk characterizes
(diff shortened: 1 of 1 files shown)


svgeesus:
Renaming this is damaging to [existing implementations](https://w3c.github.io/png/Implementation_Report_3e/#mastering), and hurts compatibility with existing content.

The sole benefit is that a small and almost theoretical class of applications (the PNG-specific editor) is told to discard it. That doesn't seem worth the breakage for a much larger class (normal image editors and content generation tools, browsers).

ProgramMax:
I agree with your points. I'll even add one: It's only for PNG editors which are unaware of the mDCv/mDCV chunk. If we could magically update all PNG editors, there would be no problem cases.

The list of existing implementations for mDCv (and cLLi) is small. I think we could easily enough reach out to them to make the change. I also think the list of content made using these tools will be exceedingly small, if not zero. Not a lot of content contains this metadata yet, to be transferred into a PNG.

(As a side note, there are rules on ordering restrictions that require mDCV/cLLI. Only PNG editors need to be aware of the rules. But it does make the spec self-inconsistent.)

There are a handful of PNG editor programs, such as optipng. There are very few of these, though. So in my mind, it is a trade-off is for those existing PNG editors (which don't know about mDCV/cLLI) vs. the few existing implementations of mDCv/cLLi. One of them has to lose/be updated. And I think it is easier for us to update the programs that have shown they are updating.

Just some points to think about. Happy to go with other thoughts. We can also talk it over in our next meeting.

svgeesus (approved):
As agreed at 25 Nov call

ProgramMax:
[MediaInfo is already landing this](https://github.com/MediaArea/MediaInfoLib/pull/2148).
I contacted someone from FFmpeg about adding the change.
I'll work on the Chromium update.

That covers everyone listed in the implementation report.

PR #474, opened 2024-10-30
Rename cLLi to cLLI

ProgramMax:
The cLLI chunk depends on image data. As a result, it is not safe to copy if that pixel data changes.

This commit renames the cLLi chunk to cLLI to reflect the unsafe-to-copy behavior.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2014,7 +2014,7 @@ with these exceptions:
 
         <tr>
           <td>
-            <a class="chunk" href="#cLLi-chunk">cLLi</a>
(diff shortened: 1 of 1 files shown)


svgeesus (approved):
As agreed at 25 Nov call

ProgramMax:
[MediaInfo is already landing this](https://github.com/MediaArea/MediaInfoLib/pull/2148).
I contacted someone from FFmpeg about adding the change.
I'll work on the Chromium update.

That covers everyone listed in the implementation report.

PR #475, opened 2024-10-30
Rename eXIf to eXIF

ProgramMax:
The eXIF chunk depends on image data. As a result, it is not safe to copy if that pixel data changes.

This commit renames the eXIf chunk to eXIF to reflect the unsafe-to-copy behavior.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1388,7 +1388,7 @@ with these exceptions:
 
           <li>Miscellaneous information: <a class="chunk" href="#11bKGD">bKGD</a>, <a class="chunk" href="#11hIST">hIST</a>,
           <a class="chunk" href="#11pHYs">pHYs</a>, <a class="chunk" href="#11sPLT">sPLT</a>, <a class="chunk" href=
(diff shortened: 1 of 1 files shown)


fintelia:
Does the EXIF chunk depend on pixel data? It describes and has information about how the pixel data was obtained, but I'm not sure that's quite the same thing

ProgramMax:
None of the Exif data depends on pixel data EXCEPT the rotation.

If a PNG editor does not understand the eXIf/eXIF chunk and updates the pixel data to something new (with no rotation), the existing Exif data might incorrectly apply a rotation to it again. Which is why it cannot be safely copied.

svgeesus:
The `eXIf` chunk has been deployed in the wild for a decade. Are we really going to make a backwards-incompatible change which affects a large body of existing images, just to suit a near-mythical software category (PNG-specific image editors, as opposed to general image editors)?

What motivation to image generators have to use the new chunk, given that support for it will be slow at the start?

PR #490, opened 2025-01-17
Update SVG figures for mDCV, cLLI

svgeesus:
The earlier find and replace had missed the figures because they are separate files.

Also corrected a typo (mDVc) to mDCV


svgeesus:
@ProgramMax @chrisn could I get a review on this trivial fix?

chrisn (approved):

ProgramMax (commented):
Oh! Right! Good catch. Thank you.
Post-facto LGTM

Issue #497, opened 2025-02-03
Missing eXIf note from the Security Considerations section


ctruta:
In the original `eXIf` chunk proposal from 2017-Jun-15 (http://ftp-osl.osuosl.org/pub/libpng/documents/proposals/eXIf/png-proposed-eXIf-chunk-2017-06-15.html) there was an addition to the Security Considerations section.

> The Exif specification does not contain a requirement that tag "value offset" pointers actually point to a valid address within the file (see Paragraph 4.6.2). Offsets are measured in bytes from the beginning of the TIFF header (the "II" or "MM"). Although this seems to be an implicit requirement, decoders should be prepared to encounter invalid pointers and deal with them appropriately.

The above paragraph is missing from the latest draft specification. I'd like to add it in, because, in my opinion, this clarification is significant.

svgeesus:
Closed by https://github.com/w3c/png/pull/498

PR #498, opened 2025-02-03
Add eXIf note to the Security Considerations section (#497)

ctruta:
Fixing issue #497, but with a minor rephrasing of the original statement _"[...] seems to be an explicit requirement"_.


svgeesus (approved):
LGTM

PR #514, opened 2025-04-21
Clarify APNG

jbowler:
The description of the PNG Development Group is at best prejorative but certainly insulting.

My edits are accurate, helpful (extra links) and not insulting to anyone.

### DIFF
--- a/Third_Edition_Explainer.md
+++ b/Third_Edition_Explainer.md
@@ -105,7 +105,7 @@ was a feature.
 
 Firefox 3 added APNG support in 2007.
 
-The PNG development group, after much discussion, rejected the APNG proposal in 2007
+The PNG development group, after <a href="https://sourceforge.net/p/png-mng/mailman/png-mng-misc/thread/3.0.6.32.20070307210904.01283690%40mail.comcast.net/#msg2437764">much discussion</a>, rejected the APNG proposal in 2007
 on the grounds that PNG should remain a static format and that APNG, unlike MNG, did
 not clearly identify itself as a separate format.
 
--- a/Third_Edition_Explainer.md
+++ b/Third_Edition_Explainer.md
@@ -105,15 +105,20 @@ was a feature.
 
 Firefox 3 added APNG support in 2007.
 
-The PNG development group, after much discussion, **rejected APNG in 2007**,
-mostly on the grounds that it was not MNG and that PNG should remain a static format.
-Two attempts to restart the discussion in the following years failed.
+The PNG development group, after much discussion, rejected the APNG proposal in 2007
+on the grounds that PNG should remain a static format and that APNG, unlike MNG, did
+not clearly identify itself as a separate format.
 
-APNG was documented [on the Mozilla wiki](https://wiki.mozilla.org/APNG_Specification)
-and browsers and other tools added a variety of patches to their own versions of libpng,
+Two attempts to revisit the decision in the following years failed.
+
+APNG is documented on the [Mozilla wiki](https://wiki.mozilla.org/APNG_Specification)
+and some browsers and other tools added a variety of patches to their own versions of libpng
 to support APNG.
 
-APNG started to be [**added to other platforms**](https://caniuse.com/apng):
+The adoption of APNG is documented by [Alexis Deveria](http://a.deveria.com/) on the "can I use"
+web page [Animated PNG (APNG)](https://caniuse.com/apng).
+
+Some specific examples of the addition of support for APNG include:
 
 - Opera 9 in 2009
 - Safari 8 in 2014
@@ -125,19 +130,19 @@ Authoring support for APNG also improved over time.
 
 ## APNG
 
-The PNG working group was chartered in 2021
-to maintain and extend the PNG specification;
-in particular, to add the widely adopted APNG extensions to the **core PNG specification**.
-This has now been done; specification links:
+The PNG working group was chartered in 2021 to maintain and extend the PNG specification specifically
+to add the widely adopted APNG extensions to the core PNG specification.
+
+This has now been achieved;
 
  - General introduction [APNG: frame-based animation](https://www.w3.org/TR/png-3/#apng-frame-based-animation)
     - ['acTL' Animation Control Chunk](https://www.w3.org/TR/png-3/#acTL-chunk)
     - ['fcTL' Frame Control Chunk](https://www.w3.org/TR/png-3/#fcTL-chunk)
     - ['fdAT' Frame Data Chunk](https://www.w3.org/TR/png-3/#fdAT-chunk)
 
-These are precisely the same as the Mozilla documentation,
+These are identical to the existing the (Mozilla)[https://www.mozilla.org/] (documentation)[https://wiki.mozilla.org/APNG_Specification],
 bringing the PNG specification into line with a
-decade of deployed content and implementations.
+decade of deployed APNG content and implementations.
 
 ## HDR support
 


ProgramMax:
FWIW, the explainer is intended for W3C members (outside the WG) who will be voting on the new standard.

It is important to highlight a change they would be voting on had previously been rejected. Not highlighting it could be seen as trying to sneak it past them.

I am okay with the changes in the first paragraph. It adds extra info to help people understand and make a decision.
The "Two attempts..." might not be clear enough, though.

For example, if I was unaware of the situation and I read "Two attempts to restart the discussion...failed" I would assume it wasn't fully revisited. But if I read "Two attempts to revisit...failed" I would assume it was fully revisited and had now fully failed three times.

jbowler:
> It is important to highlight a change they would be voting on had previously been rejected.

The problem with the explainer as written is that it draws conclusions which are interpretative and maybe partisan without presenting any evidence.  I toned it done but I didn't add the evidence since I don't know exactly what @svgeesus is referring to:

>The PNG development group, after much discussion, **rejected APNG in 2007**, mostly on the grounds that it was not MNG and that PNG should remain a static format.  Two attempts to restart the discussion in the following years failed.

That is @svgeesus 's interpretation of the **discussion** but it doesn't quote its sources so readers can't verify its correctness.

@svgeesus : the discussion is [here](https://sourceforge.net/p/png-mng/mailman/png-mng-misc/?viewmonth=200703), the CFD is easy to find (in April I believe) but I'm not sure what you meant by the two "restarts".  The discussions from March 2007 also include other proposals interleaved making it difficult to follow such as Adeluc's (competing, IRC) mPNG.

It would work to add hyperlinks to the original discussion along with links to Chris's attempted "restart[s] [of] the discussion".  I.e. enhance "rejected the APNG proposal" in my edit with an appropriate hyperlink, perhaps to the whole thread:

\+ The PNG development group, [after much discussion](https://sourceforge.net/p/png-mng/mailman/png-mng-misc/thread/3.0.6.32.20070307210904.01283690%40mail.comcast.net/#msg2437764), rejected the APNG proposal in [2007](https://sourceforge.net/p/png-mng/mailman/png-mng-misc/thread/3.0.6.32.20070406090031.0197af68%40mail.comcast.net/#msg2438582).

I found two "revote/discussion" threads and I assume these are the ones you are referring to:

o [2013](https://sourceforge.net/p/png-mng/mailman/png-mng-misc/thread/CA%2BPdXcutiut4XyfRmW4kWbbJ96sYOQxffczTU1MKck8z2w%2BYjA%40mail.gmail.com/#msg31083071)
o [2017](https://sourceforge.net/p/png-mng/mailman/png-mng-misc/thread/fa3adedb-5d2c-c892-4ae4-ef928270e542%40w3.org/#msg36039389)

To quote the initial message from 2013 verbatim:

>I have received a request, off list, to call for a revote of the APNG proposal that failed six years ago.

Note, @programmax, this is not an attempt to "_restart_ the **discussion**"; it is a request to "_revisit_ the **decision**".

The second thread was started by @svgeesus and seemed more open.  It wasn't clear what was being proposed but a discussion was started, of a sort, but never completed.  See the [last message in the thread](https://sourceforge.net/p/png-mng/mailman/png-mng-misc/thread/2046657518.192027.1511174247179%40ox.hosteurope.de/#msg36129511)

I'm trying to avoid pejorative comments; a _discussion_ was started, it wasn't an attempt, it happened!  However the _decision_ was not revisited because no CFV was initiated, there wasn't even a CFD.  My edit is consistent with that.

Third_Edition_Explainer.md
```diff
@@ -105,15 +105,20 @@ was a feature.
 
 Firefox 3 added APNG support in 2007.
 
-The PNG development group, after much discussion, **rejected APNG in 2007**,
-mostly on the grounds that it was not MNG and that PNG should remain a static format.
-Two attempts to restart the discussion in the following years failed.
+The PNG development group, after much discussion, rejected the APNG proposal in 2007
+on the grounds that PNG should remain a static format and that APNG, unlike MNG, did
+not clearly identify itself as a separate format.
```
svgeesus, line 110: 1. Automatic fallback from APNG to PNG was presented as a strength; forcing loading to fail with a unknown format is a different (and less good) design
2. PNG and APNG have separate MIME types, so the identification is taken care of

Third_Edition_Explainer.md
```diff
@@ -105,15 +105,20 @@ was a feature.
...
-and browsers and other tools added a variety of patches to their own versions of libpng,
+Two attempts to revisit the decision in the following years failed.
+
+APNG is documented on the [Mozilla wiki](https://wiki.mozilla.org/APNG_Specification)
+and some browsers and other tools added a variety of patches to their own versions of libpng
 to support APNG.
 
-APNG started to be [**added to other platforms**](https://caniuse.com/apng):
+The adoption of APNG is documented by [Alexis Deveria](http://a.deveria.com/) on the "can I use"
+web page [Animated PNG (APNG)](https://caniuse.com/apng).
```
svgeesus, line 119: CanIUse is very well known in the W3C community and does not really merit the detailed introduction. Not that I am opposed, but it really isn't needed.

ProgramMax:
Right. I think you and I are on the same page there. What I meant with my concern is a reader could pull the wrong interpretation from it. And that wrong interpretation could have drastic consequences. So I want to preemptively assist the reader.

An attempt to revisit the decision failed. Agreed. But did it fail because the decision *was* revisited and upheld? Or because it never reached the consideration stage? That is what I meant by "I would assume it was fully revisited and had now fully failed three times." It never got to the decision stage, so it wasn't rejected 3 times.

If I am a reader, I will have very different mindsets if I'm asked to approve something which was rejected 3 times vs. something that was rejected once.

I think we could reword that part further to be more clear.
Perhaps something like "Two attempts to revisit the decision (in 2013<link> and 2017<link>) were made. But they never resulted in a call for votes."

Third_Edition_Explainer.md
```diff
@@ -105,15 +105,20 @@ was a feature.
 
 Firefox 3 added APNG support in 2007.
 
-The PNG development group, after much discussion, **rejected APNG in 2007**,
-mostly on the grounds that it was not MNG and that PNG should remain a static format.
-Two attempts to restart the discussion in the following years failed.
+The PNG development group, after much discussion, rejected the APNG proposal in 2007
```
svgeesus, line 108:
```suggestion
The PNG development group, after <a href="https://sourceforge.net/p/png-mng/mailman/png-mng-misc/thread/3.0.6.32.20070307210904.01283690%40mail.comcast.net/#msg2437764">much discussion</a>, rejected the APNG proposal in 2007
```

svgeesus:
> I think we could reword that part further to be more clear.
> Perhaps something like "Two attempts to revisit the decision (in 2013 and 2017) were made. But they never resulted in a call for votes."

Right. In both cases, lack of critical mass of people interested in actually documenting the deployed reality (the few contributing to the discussion were more interested in redesigning an entirely new, not backwards compatible, format) meant that there was no point to start a call for votes.

Third_Edition_Explainer.md
jbowler: Yes @svgeeus but you didn't suggest a change.  I guess you could argue that the static vs multiple image objection was only one issue; in the original vote I raised a number of issues.  Nevertheless as I remember the discussions it was compelling and it was fixed by 2017.  You pointed this out in the _informal_ discussion yourself[1](https://sourceforge.net/p/png-mng/mailman/message/36039481/).

So what I identified in my edit as the grounds for the failed CFV had been partially fixed by 2017 (but not, apparently, by 2013) and Glenn had declared that this was sufficient for him to flip his vote[2](https://sourceforge.net/p/png-mng/mailman/message/36039417/).

The problem in 2017 was not that a discussion was not started and it was not that the proposal was rejected.  The issue was accurately summarised in the final message[3](https://sourceforge.net/p/png-mng/mailman/message/36129511/).

Glenn would probably, almost certainly, have folded on the file extension.  No one relies on file extensions when reading image files.  Glenn had already folded on the "signature" issue; the first few bytes of the file.  This is what image processing applications use when working out the format.  That was a big fold, but even if Glenn did care for some reason no one else does.  This would have worked:

\- APNG can be identified using the image/apng MIME type.
\+ APNG can be identified using the image/apng MIME type.  Where necessary to identify PNG files as containing APNG data the file extension .apng may be used however it is recommended that decoders with APNG capability recognise APNG data in any file with the PNG signature.

Other people who voted NO also seemed likely to flip.  The CFV would have passed after minimal, possibly no, CFD.  Multiple CFDs and CFVs on the same chunk with slightly different specifications are possible - that's how eXIf was done (there were two votes in parallel.)

jbowler:
> In both cases, lack of critical mass of people interested in actually documenting the deployed reality

That was a non-issue by 2017 and probably by 2013 too; see the final paragraph of my own comment here[1](https://sourceforge.net/p/png-mng/mailman/message/31083662/).  Also @svgeesus the comment you make immediately following the above doesn't really jibe with my third paragraph:

>In terms used by standards organizations rather than lawyers I think APNG has to be considered as an existing extension which is being offered for standardization.  It's not like before where issues about whether this is the right way to do animation and whether, in fact, it should be done at all were (correctly) raised.

Nevertheless, even though I at least regarded deployment as moot, a number of people did, in fact, "document the depolyed reality" in response to Greg's request[2](https://sourceforge.net/p/png-mng/mailman/message/31083754/).  My own response[3](https://sourceforge.net/p/png-mng/mailman/message/31092607/) illustrates the problem we, humble FOSS developers, have in producing such documentation; Google subsequently changed the image search API I used to obtain representative PNG images at the end of 2011 and, by my reading of the more licenses, basically removed the license to do what I had done in 2011.

@ProgramMax apparently has permission to use Google APIs and asserted previously that he was going to research actual deployments of features relevant to PNGv3 such as the eXIf chunk and its position.  I haven't seen the result of that yet but it is an extremely valuable thing to do.  It's a pity special permission is needed.  It shuts individuals like me out of the process.

Please also read this comment[4](https://sourceforge.net/p/png-mng/mailman/message/31092641/)  I'll quote the body here;

>Note that it wasn't us (myself, Stuart, or Mozilla to my knowledge) that have asked for a re-vote; I'm not sure where that's coming from.  APNG was a useful solution to a problem 6 years ago, when animated gifs were used all over the place on the web for effects and animations.  These days, web design has largely moved on, aided by declarative animations via CSS
Animations/Transitions, SVG/SMIL, etc.  APNG isn't going to replace funny cat GIFs.

That threw cold water on the whole discussion.

After that message I tried using the Google API again and came up with this data set[5](https://sourceforge.net/p/png-mng/mailman/message/31107417/)

At no point do I see any instances in 2013 of "redesigning" anything.  The problem is that no one was championing this.  This is the way FOSS works.  Someone needs to step up to the plate and do the work to move a proposal, or just a vote, through the process.

jbowler:
> I do not have special Google API access. Anyone can do it. https://chromium.googlesource.com/chromium/src/+/HEAD/docs/use_counter_wiki.md

From that page:

>Note, measurements are only recorded for HTTP/HTTPS pages. Usage on the new tab page, on data URLs or file URLs are not.

So what did you do?  The main question in this context the number of APNG images out there, so how did you do that?  (Or, indeed, the number of images with eXIf; basically the same question or query but for a different chunk.)

PR #528, opened 2025-07-07
Negative  values allowed in `cHRM`

svgeesus:


fintelia:
I think it would be good to add an informational note in the cHRM section along the lines of: "Previous versions of this specification used unsigned 4-byte integers for these values, but the change is backwards compatible because PNG restricts 4-byte unsigned integers to the range 0 - 2^31-1 (rather than 0 - 2^32-1)".

ProgramMax (commented):
LGTM

svgeesus:
So can I add another commit here even after merge or should I make a new PR to add the explanation?

chrisn:
Adding a note is a good idea, I'd suggest here as logically it's part of this change.

PR #551, opened 2025-12-26
Filtering uses corresponding bytes. Add diagrams and fix wording.

### LINKED ISSUE #543: Filtering description confussion
I feel that the description of the filtering process is described in a confusing way. The beginning of 9.2 states:

> Filters are applied to bytes, not to pixels, regardless of the bit depth or color type of the image.

However in the table right bellow it:

> _a_  the byte corresponding to x in the pixel immediately before the pixel containing x (or the byte immediately before x, when the bit depth is less than 8);

To me this  feels incredibly confusing and self contradicting. I got caught up on this wording quite a few times personally.

svgeesus:
Will close #543

### DIFF
--- a/figures/serializing-and-filtering-16bit-scanline.svg
+++ b/figures/serializing-and-filtering-16bit-scanline.svg
@@ -1,0 +1,252 @@
+<?xml version="1.0" ?>
+<svg viewBox="0 0 800 340" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+<style >
+    @font-face {
+      font-family: OS_Cond_Med;
+      src: url(./fonts/OS_Cond_Med.woff2);
+    }
+  rect {
+    stroke:none;
+    fill:white;
+    }
+  text {
+    font-family: OS_Cond_Med;
+    font-size:16px;
+    text-anchor:start
+    }
+  text.small {
+    font-size: 12px;
+  }
+  .description > text {
+    font-size: 18px;
+  }
+  line {
+    stroke:black;
+    stroke-width:2
+    }
+</style>
+<defs>
+
+<symbol id="filter1" >
+<rect width="8" height="8" style="fill:#EECCAA" />
+</symbol>
+
+
+
+
+
+<symbol id="pixelfr" >
+<rect width="8" height="8" style="fill:#EEAAAA" />
+</symbol>
+
+<symbol id="pixelfg" >
+<rect width="8" height="8" style="fill:#AAEEAA" />
+</symbol>
+
+<symbol id="pixelfb" >
+<rect width="8" height="8" style="fill:#AAAAEE" />
+</symbol>
+
+<symbol id="pixelfa" >
+<rect width="8" height="8" style="fill:#DDDDDD" />
+</symbol>
+
+
+
+<symbol id="pixelr" >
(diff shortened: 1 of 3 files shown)


svgeesus:
@ProgramMax could you review, please?

index.html
```diff
@@ -2552,7 +2552,16 @@ <h2>Filtering</h2>
...
         </object>
         <figcaption>
-          Serializing and filtering a scanline
+          Serializing and filtering a scanline, 8 bits per component
+        </figcaption>
+      </figure>
+
+      <figure id="serializing-and-filtering-scanline-16">
+        <!-- Maintain a fragment named "figure49" to preserve incoming links to it -->
+         <object id="figure49b" data="figures/serializing-and-filtering-16bit-scanline.svg" type="image/svg+xml">
```
ProgramMax, line 2560: The comment says to keep the name "figure49". But the new ID is "figure49b" (with a "b" at the end).

I think we don't need the ID nor the comment. The figure's ID should be fine.

figures/serializing-and-filtering-16bit-scanline.svg
ProgramMax, line 1: The "msb" text seems not centered.
The "lsb" text is fine.

Can you center both?

index.html
```diff
@@ -2690,10 +2700,18 @@ <h2>Filter types for filter method 0</h2>
       </table>
 
       <figure id="filter-byte-positions">
-        <object id="figure72" data="figures/filter-bytes.svg" type="image/svg+xml" style="width: 12vw">
+        <object id="figure72" data="figures/filter-bytes.svg" type="image/svg+xml" style="width: 12.5vw">
```
ProgramMax, line 2703: Although this was originally missing (and not the responsibility of your pull request), can we add:

above this line?
I feel like it should have been there and was missed earlier.

ProgramMax: Also confirming the change from width: 12vw to 12.5vw was intentional. I assume it was. But calling it out just in case.

index.html
```diff
@@ -2690,10 +2700,18 @@ <h2>Filter types for filter method 0</h2>
...
+        </object>
+        <figcaption>
+          Positions of filter bytes a, b and c relative to x, for 1 to 8 bits per component
+        </figcaption>
+      </figure>
+
+      <figure id="filter-byte-positions-16">
+        <object id="figure72" data="figures/filter-bytes-16bit.svg" type="image/svg+xml" style="width: 22.5vw">
```
ProgramMax: This duplicates id="figure72". Previously, a "b" was added to the end. We could do that here.
But I think the better option is to not add an ID, as described above.

ProgramMax (approved):
Sorry for the delay.
I have a few comments and suggest we make changes. But for the most part, it looks good to me.
Thank you for doing this.

index.html
svgeesus: It was

figures/serializing-and-filtering-16bit-scanline.svg
svgeesus: Done

index.html
svgeesus: Comment removed. ID left so someone can link to that figure if wanted.

ProgramMax: The figure itself already has an ID though, "serializing-and-filtering-scanline-16", right?

PR #554, opened 2026-01-21
Latest published version is png 3rd edition, fix #553

### LINKED ISSUE #553: "Latest published version" points to non-existent https://www.w3.org/TR/png-4/
Should probably use `png` instead of `png-4` here?
https://github.com/w3c/png/blob/67a22424d7c57446eb961f367755d3947b0aaf9f/index.html#L15

svgeesus:


ProgramMax (approved):
LGTM

PR #562, opened 2026-02-19
[Extensions] Update broken links, add missing changes for 1.6.1

svgeesus:
I noticed that the Extensions document still had the old repo name, thus giving broken links; also the /TR and History links were broken (since it has never been published to /TR)

Also the revision history had not been updated for 1.6.1


ProgramMax (approved):
LGTM

PR #570, opened 2026-05-28
Fix incorrect numbers in example 9, #567

### LINKED ISSUE #567: mDCV example wrong value
Similar to #539, this time in https://www.w3.org/TR/png-3/#example-9

The x coordinate for the Display P3 green primary is `0.265`, so `0.265 / 0.00002 = 13250`, not `13520`
The correct hex would then be `33 C2` instead of `34 D0`

Found while manually assembling a chunk in a hex editor.

svgeesus:
There was a typo in the chromaticity coordinate, then carried forward to the hex values.

Checked against the [Display P3 official definition](https://registry.color.org/rgb-registry/displayp3)

### DIFF
--- a/index.html
+++ b/index.html
@@ -4128,8 +4128,8 @@ with these exceptions:
 
               <tr>
                 <td>(0.265, 0.69)</td>
-                <td>{ 13520, 34500 }</td>
-                <td>{ 34 D0, 86 C4 }</td>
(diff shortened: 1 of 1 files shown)


chrisn (approved):

ProgramMax:
post-facto LGTM

PR #573, opened 2026-06-23
Add chunk for HDR Reference White level, fix #390

### LINKED ISSUE #390: Add PNG dWLm (Diffuse white luminance metadata ) chunk to conform to ISO 22028-5
ISO 22028-5 High dynamic range and wide colour gamut encoding for still images (HDR/WCG) identifies a very important parameter (Diffuse white luminance metadata ).

This metadata allows us to determine at which point highlight compression might occur during HDR to SDR conversion.

svgeesus:
As noted in 

 - #390 

this references the definition in the  (freely available) ITU-R BT.2409-9 specification, which becomes a normative reference, while noting that this is technically aligned with ISO 22028-5 (which costs 135CHF).

This should satisfy W3C criteria regarding normatively referencing specifications which are not freely available.

Both references are added to the bibliography, in their most recent (2026) editions.

For consistency, the luminance level is stored in exactly the same way as in the `cLLI` chunk (four bytes, value is multiplied by 10,000).

I also added the chunk to the [chunk ordering table](https://www.w3.org/TR/png-3/#5ChunkOrdering) but will leave adding it to the six SVG diagrams to another pull request, once this is merged.

Because the chunk is optional and contains only one value, I did not add language (like we have in `cLLi`) that a zero value means undefined. Just don't add the chunk, in that case.

I added an encoding example, similar to the ones for `cLLI`, for 160 nits.

**Note:** I am aware of the [earlier pull request for an earlier version of this chunk](https://github.com/w3c/png/pull/391) but there were many unresolved requests for changes and the references were also from 2023 so it was easier to start from scratch.

### DIFF
--- a/index.html
+++ b/index.html
@@ -179,18 +179,20 @@
         "ITU-R-BT.2020": {
           title: "Parameter values for ultra-high definition television systems for production and international programme exchange",
           publisher: "ITU",
+          date: "2015-10",
           href: "https://www.itu.int/rec/R-REC-BT.2020"
         },
         "ITU-R-BT.2100": {
           title: "ITU-R BT.2100, SERIES BT: BROADCASTING SERVICE (TELEVISION). Image parameter values for high dynamic range television for use in production and international programme exchange",
           publisher: "ITU",
-          date: "2018-07",
+          date: "2025-02-05",
           href: "https://www.itu.int/rec/R-REC-BT.2100"
         },
         "ITU-R-BT.2390": {
           title: "High dynamic range television for production and international programme exchange",
           publisher: "ITU",
-          href: "https://www.itu.int/dms_pub/itu-r/opb/rep/R-REP-BT.2390-11-2023-PDF-E.pdf"
+          date: "2025-03",
+          href: "https://www.itu.int/pub/R-REP-BT.2390-12-2025"
         },
         "ITU-R-BT.2408-9": {
             title: "Guidelines for operational practices in high dynamic range television production",
--- a/index.html
+++ b/index.html
@@ -4354,6 +4354,12 @@ with these exceptions:
           as is used in [[ISO_22028-5]].
         </p>
 
+        <p class="note">As HLG is a relative system, the value of 203 cd/m<sup>2</sup> 
+          is only correct when viewed under reference viewing conditions
+          on a 1000 cd/m<sup>2</sup> monitor.
+          With other screen brightnesses and under different viewing conditions, 
+          this value can be expected to change.</p>
+
         <p>Previously, it was not possible to indicate
           that the content was mastered to some other reference level,
           such as 100 cd/m<sup>2</sup>.
--- a/index.html
+++ b/index.html
@@ -2054,6 +2054,16 @@ with these exceptions:
 
         <tr>
           <td>
+            <a class="chunk" href="#hRWl-chunk">hRWl</a>
+          </td>
+          <td>No</td>
+          <td>
+            Before <a class="chunk" href="#11PLTE">PLTE</a> and <a class="chunk" href="#11IDAT">IDAT</a>.
+          </td>
+        </tr>
+
+        <tr>
+          <td>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4303,7 +4325,89 @@ <h2><span class="chunk">cLLI</span> Content Light Level Information</h2>
...
+
+        <section id="hRWl-chunk">
+            <h2><span class="chunk">hRWl</span> HDR Reference White Level</h2>
+            <p>The four-byte chunk type field contains the hexadecimal values</p>
+
+          <pre>
+68 52 57 6C
+</pre>
+
+        <p>If present, the <span class="chunk">hRWl</span> chunk identifies a characteristic of <a>HDR</a> content.It adds static metadata which is helpful when compositing a mix of HDR and SDR content.</p>
```
chrisn:
```suggestion
        <p>If present, the <span class="chunk">hRWl</span> chunk identifies a characteristic of <a>HDR</a> content. It adds static metadata which is helpful when compositing a mix of HDR and SDR content.</p>
```

chrisn (commented):
Just one very minor suggestion.

simontWork:
This corresponds to a code value of 58% in PQ and 75% in HLG on a 1000 cd/m<sup>2</sup> display.

I think there needs to be an explicit statement that when using HLG, just because the composition happened on a pseudo-1000nit monitor, the brightness at which the item will be displayed is variable.

(I would argue that it's far easier to choose a different pesudo-monitor brightness to do the transform at if you want shadows and mid-tones to match - having 1000 written here makes it seem that that's the correct option)

svgeesus:
I was drawing from this text in BT.2408-9, section **2.1 HDR Reference White**:

> The reference level, HDR Reference White, is defined in this Report as the nominal signal level obtained from an HDR camera and a 100% reflectance white card resulting in a nominal luminance of 203 cd/m2 on a PQ display or on an HLG display that has a nominal peak luminance capability of 1 000 cd/m2. That is the signal level that would result from a 100% Lambertian reflector placed at the centre of interest within a scene under controlled lighting, commonly referred to as diffuse white1. There may be brighter whites captured by the camera that are not at the centre of interest and may therefore be brighter than the HDR Reference White.

(plus the values in Table 1, which is where 58% and 75% come from)

(there is a lot more text after this, but that is the first paragraph of the definition).

And isn't 75% the code value where the gamma and the log potions join, in HLG?

I agree that for display on less-capable displays, or in brighter than reference viewing conditions, the actual measured luminance for HDR Reference White will vary due to a color re-rendering step (OOTF),

Do you have a specific change to the current proposed wording?

svgeesus:
> Just one very minor suggestion.

Fixed, thanks.

svgeesus:
@simontWork waiting for your feedback

simontWork:
Hi

Is it possible to have a note?

Note: As HLG is a relative system, the value of 203 cd/m2 is only correct when viewed under reference viewing conditions on a 1000 cd/m2 monitor.  With other screen brightnesses and under different viewing conditions, this value can be expected to change.

ProgramMax:
Should the last letter of the chunk name be capitol, indicating not-safe-to-copy?

If I edit the image, am I remastering it?
And if the editor doesn't understand the chunk it will copy the old value with the previous reference white level.

svgeesus:
> Is it possible to have a note?

Added.

svgeesus:
> Should the last letter of the chunk name be capitol, indicating not-safe-to-copy?

Indeed, that would be consistent with what we decided for `cLLI` (which was originally proposed as `cLLi`)

svgeesus:
> Should the last letter of the chunk name be capitol, indicating not-safe-to-copy?

Fixed.

ProgramMax (approved):
LGTM

simontWork:
I can't see the button to review anymore, but happy with the note addition and other changes.

chrisn (approved):
LGTM

PR #575, opened 2026-07-02
Link to infra definition of code point, fix #574

### LINKED ISSUE #574: "code points" definition in I18n glossary has gone away
@aphillips wrote in [#239](https://github.com/w3c/png/issues/239#issuecomment-1332459025):

> 1. I added links above to the I18N Glossary, which is a new effort by the I18N WG to provide convenient stable references to common character encoding terminology. This allows specs such as PNG to use I18N's preferred jargon without having to spend a ton of effort recapitulating or defining locally what is pretty dense and fiddly terminology. If you are using ReSpec you can add the directive `xref: ["i18n-glossary"]` to your config block and reference terminology using ReSpec's notation as-if these terms were defined locally.

> 4. The 79 character (code point) length limit seems fairly antiquated? The days of 80 column terminals are fairly far behind us. But I don't know what the state is for PNG implementations/tools and the 79 value might be baked into implementations? I used the term "code point" here because UTF-8 is a variable width (multibyte) encoding and I wanted to make clear that the recommended limit was in characters and not bytes, at least for the `iTXt` chunk.

These review comments were adopted, and published in PNG Third Edition.

As we work on the Fourth Edition we notice that the `code points` definition is no longer present in the I18n Glossary, which is giving us a ReSpec error

> Couldn't find "code points" in this document or other cited documents: [i18n-glossary].
>
> How to fix: [See search matches for "code points"](https://respec.org/xref/?term=code+points&types=_CONCEPT_) or [Learn about this error](https://respec.org/docs/#error-term-not-found).
>
> Occurred 1 times at:
>
>    [`<a>`](https://w3c.github.io/png/#respec-offender-no-matching-definition-found) element

Has the definition moved elsewhere, or been renamed?

svgeesus:
Changed all occurrences, not just the one linking to the glossary.

(Did not change the other use of code point, ie CICP; just the one referring to text characters)

### DIFF
--- a/index.html
+++ b/index.html
@@ -3511,7 +3511,9 @@ with these exceptions:
           </table>
 
           <p>The profile name may be any convenient name for referring to the profile. It is case-sensitive. Profile names shall
-          contain only printable Latin-1 characters and spaces (only code points 0x20-7E and 0xA1-FF are allowed). Leading,
+          contain only printable Latin-1 characters and spaces (only
+          <a href="https://infra.spec.whatwg.org/#code-point">code points</a>
+          0x20-7E and 0xA1-FF are allowed). Leading,
           trailing, and consecutive spaces are not permitted. The only compression method defined in this specification is method 0
           (<a>zlib</a> datastream with <a>deflate</a> compression, see <a href='#10CompressionOtherUses'></a>). The compression
           method entry is followed by a compressed profile that makes up the remainder of the chunk. Decompression of this
@@ -4529,7 +4531,9 @@ with these exceptions:
 
           <p>Keywords of general interest SHOULD be listed in [[PNG-EXTENSIONS]].</p>
 
-          <p>Keywords shall contain only printable Latin-1 [[ISO_8859-1]] characters and spaces; that is, only code points 0x20-7E
+          <p>Keywords shall contain only printable Latin-1 [[ISO_8859-1]] characters and spaces; that is, only
+          <a href="https://infra.spec.whatwg.org/#code-point">code points</a>
+          0x20-7E
           and 0xA1-FF are allowed. To reduce the chances for human misreading of a keyword, leading spaces, trailing spaces, and
           consecutive spaces are not permitted in keywords, nor is U+00A0 NON-BREAKING SPACE since it is visually indistinguishable
           from an ordinary space.</p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3511,7 +3511,9 @@ <h2><span class="chunk">iCCP</span> Embedded ICC profile</h2>
           </table>
 
           <p>The profile name may be any convenient name for referring to the profile. It is case-sensitive. Profile names shall
-          contain only printable Latin-1 characters and spaces (only code points 0x20-7E and 0xA1-FF are allowed). Leading,
+          contain only printable Latin-1 characters and spaces (only
+          <a href="https://infra.spec.whatwg.org/#code-point">code points</a>
+          0x20-7E and 0xA1-FF are allowed). Leading,
```
annevk: Couple of things:

1. I believe ReSpec has some way to link terms so you don't have to include the full URL everywhere.
2. Code points use the U+... syntax. 0x... are bytes.

annevk (approved):

index.html
svgeesus: > I believe ReSpec has some way to link terms so you don't have to include the full URL everywhere.

Ok, I changed it, but before I push that change:

> Code points use the U+... syntax. 0x... are bytes.

Ah, in that case code points is not the right reference. The byte for a **printable Latin 1 character** is not the same as the bytes for the same character from **C1 Controls and Latin-1 Supplement**. And this text is about the allowed byte values.

I know, if we were designing PNG today all the text would be UTF-8 but that wasn't the case in 1994.

So really this should be "bytes" which would link to [bytes](https://infra.spec.whatwg.org/#bytes).

annevk: Then you should reference https://infra.spec.whatwg.org/#byte probably.

aphillips: Add "infra" to the list in "xrefs" in the ReSpec configuration to get auto-linking.

> So really this should be "byte values".

The problem here is that, yes, these are bytes, but we don't just mean random bytes. They are _explicitly_ in the Latin-1 (ISO 8859-1) character encoding. (Note that the Encoding spec defines Latin-1 as windows-1252 for a variety of good reasons not important here.) I don't really want to bring in isomorphic encode/decode here.

Perhaps:

> contain only [=bytes=] in the ranges 0x20-7E and 0xA1-FF, representing the 
> corresponding printable [=code points=] in the Latin-1 [=character encoding=] (or the space character).
> Leading,

annevk: Right, you could also talk about code points and then use https://infra.spec.whatwg.org/#isomorphic-encode over the resulting string to get a byte sequence.

PR #578, opened 2026-07-28
Fix rare incorrect chunk ID spellings (`iTXT`, `tEXT`, `zTXT`, `gIFT`, `eXIF`)

generalmimon:
I found these errors in chunk ID casing using [ripgrep](https://github.com/BurntSushi/ripgrep) as follows (the output corresponds to https://github.com/w3c/png/commit/426e411b13ae3853984f14c9269766adec8d6d15, i.e. the latest commit on [`main`](https://github.com/w3c/png/tree/main) as of this writing):

```console
$ rg --version
ripgrep 15.2.0 (rev e89fff89ac)

features:+pcre2
simd(compile):+SSE2,-SSSE3,-AVX2
simd(runtime):+SSE2,+SSSE3,+AVX2

PCRE2 10.45 is available (JIT is available)

$ rg --no-filename -o -w '[A-Za-z]{2}[A-Z][A-Za-z]' | sort -u | uniq -i -D
cICP
CICP
cLLi
cLLI
eXIf
eXIF
EXIF
fdAT
fDAT
gIFt
gIFT
iTXt
iTXT
sRGB
SRGB
tEXt
tEXT
zTXt
zTXT
```

I searched for each of these throughout the entire repository, and for most of them, I concluded that they were fine and shouldn't be changed. I corrected only the remaining ones.

Note that b5e58c752fc43397d9b0c029ecbadb07883a2ee4 and cabfbfa0ecc4bf4ec2621d95cabf63de2ee08ec0 modify the source file `extensions/Overview.bs`, but I deliberately didn't update `extensions/Overview.html`, because I realized that you would probably want to regenerate it rather than modifying it directly.

I don't know how to regenerate `extensions/Overview.html`, and it would probably be much easier for you to do it the way you want.

### DIFF
--- a/extensions/Overview.bs
+++ b/extensions/Overview.bs
@@ -1163,7 +1163,7 @@ What are <code>x0</code> and <code>x1</code> for?
     <ul>
 
     <li>Added <code>pCAL</code> chunk and related sample code
-    </li><li>Deprecated the <code>gIFT</code> chunk
+    </li><li>Deprecated the <code>gIFt</code> chunk
     </li><li>Added sample gamma-correction code that uses integer arithmetic
 
     </li></ul>
--- a/extensions/Overview.bs
+++ b/extensions/Overview.bs
@@ -1098,7 +1098,7 @@ What are <code>x0</code> and <code>x1</code> for?
     <li>24 Jan Dec 2022 (version 1.6.0)</li>
 
     <ul>
-        <li>Moved <span class="chunk">eXIF</span> to main PNG document</li>
+        <li>Moved <span class="chunk">eXIf</span> to main PNG document</li>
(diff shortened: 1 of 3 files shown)


svgeesus (approved):

generalmimon:
@svgeesus Thank you for merging the PR so quickly!

I'd just like to note that [`extensions/Overview.html`](https://github.com/w3c/png/blob/5bf360a93222038ace381d72152f379caa9863ab/extensions/Overview.html) should be regenerated with Bikeshed at some point, but it's probably not necessary to do that right now. If you think it can wait until someone makes a more substantial change to `extensions/Overview.bs`, I don't mind.

